PR #7, opened 2015-02-03
fix(show): Make sure this hide overrides everything

PWKad:


PWKad:
Placed it outside on accident

EisenbergEffect:
I added it anyways. It's probably a good idea to have that, don't you think?

EisenbergEffect:
Doh. I didn't even see that. I'll get our other one in now...

PR #12, opened 2015-02-07
feat(repeat): implement support for map

martingust:
This one I have iterated over a few times along the way as understood more how its working. The last I came up with was to just handle the change records for maps, switch over record.type and add/remove views accordingly. 

One difference from array is that we don't have the index, only key. Right now I'm iterating over viewSlot children to see where keys match to get the index. We might wanna come up with a better way to handle this.

The testing I have done it seems to work, but I'm probably missing something. Let me know if you see any issues/improvements to this solution and ill make necessary changes.

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -1,4 +1,4 @@
-import {ObserverLocator, calcSplices} from 'aurelia-binding';
+import {ObserverLocator, calcSplices, getChangeRecords} from 'aurelia-binding';
 import {Behavior, BoundViewFactory, ViewSlot} from 'aurelia-templating';
 
 export class Repeat {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Ok. See comments. Also, we need to not use for...of loops for now. I'd love for you to continue to work on this. Lots more work needs to be done on the repeater in general. So, review the notes int he code and lets see if we can do some improvements.

Also, I noticed that the map observer subscriptions weren't being disposed like the array subscriptions, so we need to add that.

When you get a chance, get in touch with me to let me know you are good to go for a second round of improvements on this. It's super important and I'd love to have you keep working on it!

martingust:
Awesome, thanks for feedback. I had a feeling I was abusing something. Your comments should be resolved now, I'll hit you up on Gitter...

PR #19, opened 2015-02-17
feat(templating-resources): add innerHTML attached behavior

AshleyGrant:
Initial implementation of the behavior. Ready for comments.

I implemented it per these instructions from @EisenbergEffect :
```
1. You can go ahead and implement it using the options syntax. Name the behavior inner-html.
2. So, it should be used like this: inner-html="value.bind: some.prop; sanitizer.call: mySanitizationFunction”
3. The sanitizer should be optional. If it isn’t provided it should fallback to a default santizer.
4. The default sanitizer should also be configurable.
5. Hopefully we can also support the short form: inner-html.bind=“some.prop". I’m not sure if that will work with options configured or not. I might need to do a fix for that, but you can get started on the work and I can try to fix that later.
```

This implementation addresses items 1, 2, and 3. I will need further guidance on 4.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,9 +4,10 @@ import {Repeat} from './repeat';
 import {Show} from './show';
 import {SelectedItem} from './selected-item';
 import {GlobalBehavior} from './global-behavior';
+import {InnerHTML} from './inner-html';
 
 function install(aurelia){
-  aurelia.withResources([Show, If, Repeat, Compose, SelectedItem, GlobalBehavior]);
+  aurelia.withResources([Show, If, Repeat, Compose, SelectedItem, GlobalBehavior, InnerHTML]);
 }
 
 export {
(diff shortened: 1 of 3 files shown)


PWKad:
Awesome work!  Wonder if we can get anyone with some domain expertise in injection to write some tests that go even deeper.  (they look good for now, but I know there are people with 100x more expertise than me on these subjects :) )

sorenhoyer:
I've faced a scenario where I have other binds inside the html I'm injecting with the innerHTML bind. Bindings in this injected HTML doesn't work, so would it perhaps be possible to invoke all the innerHTML bindings before the rest?

stsje:
I don't understand why you are adding a sanitizer to it. There are just too many ways to do XSS, and implementing your own sanitizer is probably not a good idea.

People come up with the weirdest ways to exploit svg, object, img tags
```
<a href="javascript:alert('hello')">Link</a>
<img src='#' onerror=alert('hello')>
```
https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

PR #20, opened 2015-02-17
fix (templating-resource): case sensitive jquery plugin names

Kennethtruyers:
Enumerate all jQuery methods when a jQuery module is not found to fix case mismatch due to html attributes being returned in lower case. Emit a warning to include it in the global behavior

https://github.com/aurelia/framework/issues/47

Note: you need to add the aurelia-logging dependency in the application config.js


EisenbergEffect:
Nice work! I'll get it merged in today most likely.

PR #22, opened 2015-02-18
feat(with): Add initial with binding

### LINKED ISSUE #16: Support "with" binding similar to knockout
In Durandal 2.0 it was possible to expose a view model as a property of a parent view model by using knockout's 'with' binding and state its view.  To aid in portability of existing Durandal apps and to support the particular use case this type of composing is needed.  In the example below "dateRangeVM" is the view model property.  By using the compose binding, a custom view could be specified.
```
<!--Date Range Selection-->
<div class="form-group" data-bind="with: dateRangeVM">
         <div data-bind="compose:{ view: 'dateRangeSelection'}"></div>
</div>
```

There may be a better way to accomplish this in Aurelia with custom elements and behaviors.  The outer view model would need to still have access to the inner view model to query state and values.

PWKad:
Trying to fix #16 
Need an idea for how to use better - 

<div with.bind="parentProp"><span>${childProp}</span></div>
Does not work - 
parentProp.childProp = 'red' 
This appears to -
parentProp = { childProp = 'blue' }


EisenbergEffect:
@PWKad Can you rebase this on master?

badunk:
oh man, I wished this was documented :).  Struggled with finding this beautiful piece of functionality I missed from Durandal for a couple days.  I thought that compose would do that automatically for me (ie, accept a view model object like in durandal), then I thought I should've done import...etc..

To me, the with binding was like the most important part of Durandal - the ability to compose view models within view models and have my views reflect that using cascading with's.

PWKad:
:)  Thanks for the feedback, this is still a work in progress which is why it isn't documented well yet.  As a note when composing a viewmodel within a viewmodel it is possible to simulate a `with` just by using the model property that is available to bind on `compose` for now.

PR #23, opened 2015-02-18
feat(with): Add WIP with binding

PWKad:
This binding should be similar to other libs where it sets context of an element

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 import {Compose} from './compose';
 import {If} from './if';
+import {With} from './with';
 import {Repeat} from './repeat';
 import {Show} from './show';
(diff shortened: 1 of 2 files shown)


PWKad:
You can check this out with 

`<div with.bind="parentValue"><span>${childValue}</span></div>`

And in the vm - 

`this.parentValue = { childValue: "blue" }`

When you set the value later with

`this.parentValue.childValue = "red"`

It doesn't update the child but when you use

`this.parentValue = { childValue: 'red' }`

It does, and properly it seems.

EisenbergEffect:
I went ahead and merged it and it will go out in the next release. We can call this an "experimental" feature while we work out the bugs.

Issue #27, opened 2015-03-11
When add an item into an array, which is bound to 'repeat' behavior, the '$parent' is undefined.


boogunote:
`context.$parent = this.executionContext;` in `function updateExecutionContext()` should be moved to `function createBaseExecutionContext()` and `function createBaseExecutionKvpContext()`.

EisenbergEffect:
@martingust Can you look into this issue please? Let me know what you find and what you think we should do. Thanks!

boogunote:
@EisenbergEffect , I have fix it. Could I push the PR? I already signed your agreement.

EisenbergEffect:
Feel free to go ahead and send the PR. I'll have @martingust review it since he's familiar with that code and then if it all looks good I can merge and get it in for our next release.

martingust:
@boogunote Can you please provide some more info on how to reproduce this? I'm not able to.

boogunote:
@martingust , Please checkout 
`https://github.com/boogunote/bn5/blob/master/src/tree-node.js` and 
`https://github.com/boogunote/bn5/blob/master/src/tree-node.html`.

I initialize view model in Ln.5 of tree-node.html.
```
  <section repeat.for="child of node.children">
    <compose
      model.bind="{'node':child, 'parentVM':$parent}"
      view-model="tree-node">
    </compose>
  </section>
```
Add check the model in `activate()` at Ln. 14 of tree-node.js.
```
  activate(model){
    console.log("TreeNode activate");
    console.log(model)
    this.node = model.node;
    this.parentVM = model.parentVM;
    this.treeVM = model.parentVM.treeVM;
  }
```


In the first round, everything work will. When `children` bind to `repeat`, I can access corrent `$parent` by `model.parentVM` in `activate()`.

But when I add an item to `children` programmatically, by `addChild() `, I check `model.parentVM` in `activate()` again and see it's `undefined`.

martingust:
Thank you. I have got it reproduced and confirmed your fix works. Will let Rob know.

martingust:
@EisenbergEffect We can go ahead and close this.

PR #28, opened 2015-03-11
init 'context.$parent' when create base execution context.

boogunote:
BUG=https://github.com/aurelia/templating-resources/issues/27


EisenbergEffect:
@martingust Can you review this?

martingust:
@EisenbergEffect I've got the issue reproduced and confirmed that the fix works. This was a use case I have not thought of, you might want to look at it as well.

PR #32, opened 2015-04-04
Repeat performance

martingust:
This should make the handleSplices-method a lot faster for all use cases I have tested, especially in cases where the array is swapped and where a lot of views are removed and added. 

Let me know if I have missed something or something does not make sense.

On line 186 is a code path to handle `Array.prototype.reverse()` it's probably the fastest way to handle it but not sure if that could be a code path for other use cases than reverse that would not work (not that I have found though).

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -180,12 +180,12 @@ export class Repeat {
       removeDelta = 0,
       view, i, ii, j, jj, row, splice,
       addIndex, end, itemsLeftToAdd,
-      removed, model;
+      removed, model, totalAdded, totalRemoved;
 
     // TODO better way to handle Array.prototype.reverse()? Could this be a path for other use cases than reverse?
     if(splices.length === 2){
-      var totalAdded = splices[0].addedCount + splices[1].addedCount;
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Ready to go? or did you want to add anything else? Did you test in both Chrome and non-Chrome? Chrome has native Array.observe and everything else has our patched implementation. Just want to make sure things work well in both scenarios.

martingust:
It works fine in Chrome, FF and Safari.

I want to handle the array reverse in a different way, because a scenario where you add N items and remove N would end up in the same code path. I will still work but probably not optimal when large list of items and low N. I'll try take a look at it today, if not, I'll be on it tomorrow.

I guess you can go a head and merge this in, and I'll just create a new PR for the improvement.

martingust:
Actually, hold off on merging this. There's an issue with the execution context index when dealing with multiple splices. Will look more inte this tomorrow.

martingust:
@EisenbergEffect Now execution context should be updated correctly and this can be merged.

PR #33, opened 2015-04-08
fix(GlobalBehavior): throw AggregateError

cmichaelgraham:
fixed incorrect passing of inner error to Error constructor

note: at some point the AggregateError function should be renamed to aggregateError, and all calls updated also.  AggregateError is a function that returns an aggregate error, not a class.  a common mistake associated with the incorrect name is the incorrect use of new when calling the function

### DIFF
--- a/src/global-behavior.js
+++ b/src/global-behavior.js
@@ -1,5 +1,6 @@
 import {inject} from 'aurelia-dependency-injection';
 import {customAttribute,dynamicOptions} from 'aurelia-templating';
+import {AggregateError} from 'aurelia-logging';
 import * as LogManager from 'aurelia-logging';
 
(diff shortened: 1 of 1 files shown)


Alxandr:
You might even want to call it `createAggregateError` or similar, so as to leave no doubts as to it's usage.

EisenbergEffect:
So, this is, of course, perfectly valid JavaScript. The reason I've done it this way is that I would like to inherit AggregateError from Error, but you cannot do that without native class support. So, I've made it look like a class and wanted to use it like a class, so that when I can change it to an actually class....I don't have to have a major breaking change in the library.

It's probably not that big of a deal for it to just be a helper method though. I will consider it.

PR #46, opened 2015-04-29
fix(GlobalBehavior): improve handler-not-found message

jdanyow:
Fixes aurelia/binding#85


jdanyow:
The highlighted items have been added to the message:

new message

PR #48, opened 2015-05-01
feat(docs): Added initial YuiDocs

PWKad:
- Added to main src files
- Added to decorators


PWKad:
@EisenbergEffect Please let me know if you see any overall formatting / syntax issues that need to be corrected

EisenbergEffect:
Ok, lets wait just a bit and see what @bryanrsmith or you come up with in terms of JSDoc. If we can get what we want out of that, then we should go that direction.

EisenbergEffect:
Github is saying this can't be merged. I'm going to close this for now until we figure oute exactly what we want to do. I imagine we'll use the Type Doc stuff if we move to TypeScript, which I also think is likely.

PR #59, opened 2015-05-14
Typo in the "if"

Mordred:
I think there is typo in the `if`. It should be `isBound` not `bound` https://github.com/aurelia/templating/blob/master/src/view.js#L14


EisenbergEffect:
Good catch! Thanks Martin!

PR #60, opened 2015-05-16
feat(all): add focus attached behavior

manuel-guilbault:


martingust:
@EisenbergEffect Can you please confirm the CLA has been signed?

src/focus.js
```diff
@@ -0,0 +1,37 @@
...
+var afterAttachedSetFocusDelay = 50;
+
+@customAttribute('focus')
+@inject(Element)
+export class Focus {
+    constructor(element) {
+        this.element = element;
+
+        this.focusListener = (e => this.value = true).bind(this);
+        this.blurListener = (e => this.value = false).bind(this);
```
bryanrsmith: the `bind`s should be unnecessary here. The arrow functions already capture the context.

bryanrsmith: Since DOM libraries like jQuery use a `false` return value to stop event propagation, I think it would clarify intent to avoid returning the assigned value.

src/focus.js
```diff
@@ -0,0 +1,37 @@
...
+            this.element.blur();
+        }
+    }
+
+    attached() {
+        this.element.addEventListener('focus', this.focusListener);
+        this.element.addEventListener('blur', this.blurListener);
+
+        if (this.value) {
+            setTimeout(() => this.element.focus(), afterAttachedSetFocusDelay);
```
bryanrsmith: If the value is changed within `afterAttachedSetFocusDelay` of attaching then this will incorrectly focus the element. I think we need to either store the timeout ID and cancel it in `valueChanged`, or check the value again inside the callback.

martingust:
Right now if an input field uses the attribute it will loose focus if you switch to another browser tab and then back again. I think it would be nice if the attribute could handle that correctly. Maybe check for activeElement in valueChanged?

src/focus.js
manuel-guilbault: You're right, the ```bind```s are useless.
For not returning the assigned value, you mean with something like this:
```this.focusListener = e => { this.value = true; };``` ?

src/focus.js
manuel-guilbault: Good catch. I'll update accordingly.

manuel-guilbault:
@martingust ok I'll check this out.
@martingust @bryanrsmith Thanks for the great review!

src/focus.js
bryanrsmith: That is what I meant, but on second though maybe they should just be methods?

EisenbergEffect:
@martingust @bryanrsmith @manuel-guilbault Glad to see you guys collaborating here. Where are we at on this? Should we put it in for tomorrow/Friday? or wait until a subsequent release.

manuel-guilbault:
Well there's still some work needed on it, I planned to do it over the weekend.

EisenbergEffect:
Sounds great. We can get it out in the next release pretty easily when it's ready. @manuel-guilbault Thanks for contributing. This will be nice!

src/focus.js
manuel-guilbault: I tried making the event handlers methods: ```onFocus(e) { this.value = true; }```.

If I attach them this way: ```this.element.addEventListener('focus', this.onFocus);``` then when the event is fired, the ```this``` keyword refers to the HTML element itself, not the ```Focus``` instance, so it doesn't work.

If I attach them this way: ```this.element.addEventListener('focus', e => this.onFocus(e));``` then when doing the detach ```this.element.removeEventListener('focus', e => this.onFocus(e));``` the handler is not correctly removed. I think this is because of the es5to6 transpiler, which creates a ```_this2``` variable in ```attached()``` and a ```_this3``` variable in ```detached()```, but I'm not 100% sure.

So unless there's another solution which I'm not seeing, I think that how I did it is the simplest working solution :)

manuel-guilbault:
Do you guys prefer having a second commit with my fixes after the review, or would you rather have me rebasing my commits together so there's only one commit in the final history?

EisenbergEffect:
If you don't mind rebasing, that would be great!

EisenbergEffect:
@manuel-guilbault This is coming along nice. I wonder if you can try something for me. Can you replace the setTimeout with a call to the TaskQueue's queueMicroTask API? I think that might be what you are after, but I'm not sure. If that doesn't work, try queueTask instead.

manuel-guilbault:
@EisenbergEffect it works well using a micro task.
I included another small change, so that if, for the same element, the show behavior is bound to the same target as the focus behavior, when the target changes to true, the focus is given after the element becomes visible through the task queue.

EisenbergEffect:
@manuel-guilbault Very nice! Do you have any interest in writing a blog post on building this? We could post it to the official blog and feature you there. It would be a great opportunity to help the community learn and we'd love to feature contributing community members too. If not, no big deal. I just want to put that out there in case you are interested.

manuel-guilbault:
@EisenbergEffect yeah I'm interested! What kind of post do you have in mind?

EisenbergEffect:
@manuel-guilbault Something that would basically be a "how to" sort of post, walking developers though what you wanted to accomplish, why and how you did it. We would also be happy to have a short 2-3 sentence bit about you at the beginning of the post. You could write that separately and then I will use that to introduce you. If you want to include links to github or other ways to contact you, we would be happy to include those too.

manuel-guilbault:
@EisenbergEffect I'll start working on that. What is the language/notation used on the blog (markdown, HTML, etc)?

jdanyow:
@manuel-guilbault  :+1:  the blog uses markdown

manuel-guilbault:
@jdanyow ok thanks! I'll keep you guys posted on my progress

EisenbergEffect:
@manuel-guilbault Ok, looks good to me. Are you satisfied or did you need to make any more changes?

manuel-guilbault:
@EisenbergEffect the only thing I wanted to add was related to my previous PR (add default binding mode for custom attribute), so the focus attribute can be two-way bound by default, but this can't happen before the next release of the templating module. How do you want to proceed on this? You can either wait for the next templating release before merging this PR, so I can add the default binding mode, or you can merge and I'll send another PR after the release. As you wish :)

EisenbergEffect:
I'll go ahead and merge it shortly. We don't have to wait for the next release to fix it though. If you want, to send another PR with the fix, I can test it against the master of templating. These will both release/update together this week.

manuel-guilbault:
@EisenbergEffect I've pretty much finished the blog post. I've got it on a [gist](https://gist.github.com/manuel-guilbault/0ef2923baa8be4ffac6b). Let me know what you think.

manuel-guilbault:
@EisenbergEffect ok will do

EisenbergEffect:
Ok. This is merged in. If you want to send another PR to add the default binding mode, that would be great. I will review your article and then maybe we can plan to run it next week? or the week after? I need to see how it fits into the schedule.

manuel-guilbault:
@EisenbergEffect ok I'll send another PR as soon as I can (probably sometime in the next few days). For the article, it's as you wish.

PR #65, opened 2015-06-01
fix(repeat): unbind children on property change

jdanyow:
@martingust can you review this?  I'm a little out of my element here.  I was seeing a memory leak due to the repeat's child views never unbinding when the property the repeat was bound to was assigned a new array instance.

Here's an example:

```javascript
export class Foo {
  bars = [];  // bind a repeat to this.
  constructor() {
    setInterval(() => {
      this.bars = ['a', 'b', 'c'];  // this should cause the repeat's child views to unbind
    }, 0);
  }
}
```

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -93,11 +93,17 @@ export class Repeat {
 
   processItems() {
     var items = this.items,
-      viewSlot = this.viewSlot;
+      viewSlot = this.viewSlot,
+      views, i;
 
     if (this.disposeSubscription) {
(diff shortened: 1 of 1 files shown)


martingust:
@jdanyow Good catch. Looks good to me.

jdanyow:
Awesome- thanks!  Would you mind moving this to the "done" pipeline- I don't have permission on this repo.

EisenbergEffect:
Give me a sec and I'll merge it in.

PR #68, opened 2015-06-03
fix(if): fix passing context binding to the view factory

### LINKED ISSUE #56: TypeError when repeat is shown for first time using if.bind
I've managed to reproduce this in skeleton, which can be seen here:

https://github.com/CharlesPockert/skeleton-navigation/tree/master/src

Basically, I've created a custom element which is a collapsible panel and I was messing about with showing/hiding it using `if.bind`. I found that If I nest a `repeat.for` inside an element and use an `if.bind` attribute on it, when the bound property is truthy and it previously was not, the element throws a TypeError:

`Uncaught TypeError: Cannot read property '__observers__' of undefined`

In the demo app I've simply provided an array of ints to repeat over and a button to toggle the bound property (`showItems`). Click the button to see the error.

This appears to work fine if I use show.bind (I assume the elements are already in the DOM with `show`).

It breaks if I use `if.bind` on any parent element.

Mordred:
Closes #56 

This bug is blocking me, so I made a PR. Can you review it?

### DIFF
--- a/src/if.js
+++ b/src/if.js
@@ -20,6 +20,12 @@ export class If {
     this.showing = false;
   }
 
+  bind(executionContext) {
+    // Store parent executionContext, so we can pass it down
+    this.executionContext = executionContext;
+    this.valueChanged(this.value);
+  }
+
   valueChanged(newValue){
     if (!newValue) {
       if(this.view){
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Looks good. I'll get it in tonight and it will go out in the next release, probably Friday.

PR #71, opened 2015-06-09
fix(repeat): remove views when bindning new items

### LINKED ISSUE #69: bug using repeat.for inside with block
Consider the following scenario:

A ViewModel like:

    export class ViewModel
    {
       submit()
       {
            this.child = {subchild: { items: ["a", "b"]}}
       }

       activate()
       {
           this.child = { subchild: { items: ["a", "b"]}}

       }
    }

bound to the following view:

    <template>
       <section>
            <form role="form" submit.delegate="submit()">
           <ul with.bind="child">
                  <li repeat.for="item of subchild.items">${item}</li>
           </ul>
            <button type="submit" class="btn btn-default">Submit</button>
           </form>
      </section>
    </template>

When you press the "submit" button, the `<li>` elements get duplicated, as if the `repeat.for` binding did not clear the old elements.

I can post a repro project based on a freshly-generated skeleton if you need one.

martingust:
Tests are temporary ignored until we get the infrastructure in place to test classes with bindable property initializers. See issue aurelia/binding#90

Fixes #69

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -42,7 +42,7 @@ export class Repeat {
 
     if(!items){
       if(this.oldItems){
-        this.viewSlot.removeAll();
+        this.removeAll();
       }
 
       return;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Fixed and released.

PR #73, opened 2015-06-11
fix(repeat): preserve lifecycle when re-using views

martingust:
@EisenbergEffect Can you please review if this looks like expected behavior. When the view gets re-used the only thing that updates is the execution context's "item", one improvement might be to only update that and manually call `unbind` from inside the repeater, oppose the create a new full execution context. Let me know what you think.

@stoffeastrom I have tested this with some basic scenarios but I know you sortable plugin is a little more edge case, if you get any chance to test it out please let me know. I've created a gist with the compiled version of `repeat.js` https://gist.github.com/martingust/44ebf1f795628a229741

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -214,22 +214,25 @@ export class Repeat {
       spliceIndexLow, viewOrPromise, view,
       i, ii, j, jj, row, splice,
       addIndex, end, itemsLeftToAdd,
-      removed, model, children, length;
+      removed, model, children, length, context, spliceIndex;
 
     for (i = 0, ii = splices.length; i < ii; ++i) {
       splice = splices[i];
-      addIndex = splice.index;
+      addIndex = spliceIndex = splice.index;
       itemsLeftToAdd = splice.addedCount;
       end = splice.index + splice.addedCount;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
I think this looks pretty good. Can you perform an experiment for me? Can you make sure this works correctly when the item template has a compose element in it and the compose element's view-model is data-bound so that it renders different view-model/views based on a property of the data?

```
<li repeat.for="widget of widgets">
  <compose view-model="${widget.type}"></compose>
</li>
```

Just want to make sure that still works correctly when re-using views.

stoffeastrom:
I just copy pasted your gist @martingust without updating Aurelia ( will try that also ). However I still experience problems with $index property. It feels like it's wrongly updated. I will look into it further.

stoffeastrom:
Updated to latest and first impression is that it works much better now. Even without adding you gist @martingust .

I will play around some more and see if I can find anything.

martingust:
@EisenbergEffect Just tested and it seems to behave as expected. 

@stoffeastrom Great. Let me know if you find any issues. 

If nothing else this should be ready to be merged.

PR #76, opened 2015-06-17
added an au-src attribute

capaj:
this is not ready to be merged, I would love to add some basic test at least. Unfortunately I can't seem to run them on my machine, will try some linux machine soon. 
Another matter is to make this behavior global as are all the other behaviors here in this folder. Where is that configured?

### DIFF
--- a/src/au-src.js
+++ b/src/au-src.js
@@ -1,0 +1,27 @@
+import {inject, noView, customAttribute} from 'aurelia-framework';
+
+/**
+ * Binding to src on an image is troubling, because browsers fetch images immediatelly. To get around this, we introduce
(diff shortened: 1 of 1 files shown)


Aaike:
i also experienced this when doing `src="${image}"`
i have been avoiding this problem by using `src.bind="image"`

capaj:
@Aaike I did not know that could work. Then it is probably not needed. It would be nice to mention it in the docs somewhere, so that people are not falling into the same issue.

jdanyow:
@EisenbergEffect can you re-open this one?  We can handle this like the textarea interpolation:

https://github.com/aurelia/templating-binding/blob/master/src/binding-language.js#L205

jdanyow:
whoops- thought this was an issue not a pull- my bad

PR #78, opened 2015-06-18
fix(show): use bind hook to handle undefined values

jdanyow:
Fixes aurelia/templating#100

The valueChanged hook isn't called during initial binding when the model value is undefined because it matches the behavior property's intitial value (undefined).  Introducing the bind hook allows us to handle any initial value.

### DIFF
--- a/src/show.js
+++ b/src/show.js
@@ -32,4 +32,8 @@ export class Show {
       this.element.classList.add('aurelia-hide');
     }
   }
+
+  bind(executionContext) {
+    this.valueChanged(this.value);
+  }
 }


jdanyow:
works in these scenarios:

```html
<template>
  <div show.bind="foo">hello world 1</div>
  <div show.bind="bar">hello world 2</div>
  <div repeat.for="item of model">
    <div show.bind="item.test">${item.title}</div>
  </div>
</template>
```
```javascript
export class App
{
  model = [
    {test:0,title:"1"},
    {test:null,title:"2"},
    {test:undefined,title:"3"},
    {test:1,title:"4"},
    {test:null,title:"5"},
    {test:null,title:"6"}];

  foo = undefined;

  constructor() {
    setTimeout(()=> {
      this.foo = true;
      this.bar = true;
      this.model[5].test = true;
    }, 3000);
  }
}
```

Aaike:
sweet, thx @jdanyow !

PR #89, opened 2015-08-03
fix(if): delay call to unbind via micro task

martingust:
And adding and fixing tests for `repeat` and `if`.


EisenbergEffect:
You are THE MAN!!! Thanks, especially for the test cases.

PR #106, opened 2015-09-07
Much faster CSS fix up matcher

### LINKED ISSUE #105: CSS loader and 404 CSS font URLs
Hi there,

I updated to the latest Aurelia dist. I'm using `<require from="bootstrap.css"></require>` inside my views. There seems to be a major issue with the CSS loader and font URLs.

Bootstrap and others usually use relative URLs for loading glyph resources. It turns out, using the latest distribution of Aurelia, I'm getting 404s for font resources.

For example, what I'm getting now:
```
http://localhost:11865/fonts/fontawesome-webfont.woff2?v=4.4.0
```
What I should be getting:
```
http://localhost:11865/pkg/github/lipis/bootstrap-social@4.10.0/assets/fonts/fontawesome-webfont.woff2?v=4.4.0
```

I think the problem stems from the CSS loader using `injectStyles` which loads the CSS inline. Ex:
```
<html>
    <head>
          <style>CSS content here... using url(../font) relative</style>
    </head>
</html>
```
Inline CSS URLs would be relative to the document in which the Aurelia app is mounted from; which in this case leads to the wrong relative path.

Instead, if we used:
```
<html>
    <head>
          <link rel="stylesheet" href="path to css on server" />
    </head>
</html>
```
**Link** will respect the *relative-ness* of the path defined inside the CSS.

Thoughts?

bchavez:
Better fixes #105 with virtually no impact on performance.

The solution here is use a combination of RegEx and Code to fix up CSS URLs. Massive reduction in RegEx complexity allows us to get to where we need to be faster :zap:; then just handle the edge cases for quoted URLs. These changes have virtually no impact in loading performance now. :+1: Roughly a **3x** speedup. Here is the performance difference:

#### Before :turtle:  
screen457

#### After :zap: 
screen458

### DIFF
--- a/src/css-resource.js
+++ b/src/css-resource.js
@@ -4,10 +4,13 @@ import {Loader} from 'aurelia-loader';
 import {Container} from 'aurelia-dependency-injection';
 import {relativeToFile} from 'aurelia-path';
 
-let cssUrlMatcher = /url\(\s*[\'"]?(([^\\\\\'", \(\)]*(\\\\.)?)+)[\'"]?\s*\)/gi;
+let cssUrlMatcher = /url\((?!['"]data)([^)]+)\)/gi;
 
 function fixupCSSUrls(address, css) {
   return css.replace(cssUrlMatcher, (match, p1) => {
+    let quote = p1.charAt(0);
+    if(quote == '\'' || quote == '"')
+        p1 = p1.substr(1, p1.length - 2);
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
This makes me so very happy :) Give me a minute and I'll get it merged and a new release out.

EisenbergEffect:
Ok, it's in. You can get 0.15.2 with `jspm update`.  Thanks again!

PR #116, opened 2015-10-09
refactor(PAL): use PAL

jdanyow:
tests don't run- added `import {initialize} from 'aurelia-pal-browser'; initialize();` - didn't help.  I must be putting it in the wrong spot


EisenbergEffect:
I'm right in the middle of working on this. I'm going to close this out since the fixes will be pushed soon.

Issue #120, opened 2015-10-15
repeat.for bound to a map within an if.bind throws exception when re-binding


DarienFord:
The ```repeat.for``` binds correctly the first time, but if you are re-binding, for example if you are selecting a different item from a list, the follow exception is thrown the second time the ```if.bind``` evaluates to true:
```
Uncaught TypeError: Cannot set property '$index' of null
updateBindingContext @ repeat.js:268
handleMapChangeRecords @ repeat.js:392
bind @ repeat.js:81
bind @ aurelia-templating.js:2706
bind @ aurelia-templating.js:802
bind @ aurelia-templating.js:2710
bind @ aurelia-templating.js:802
valueChanged @ if.js:55
BindableProperty.createObserver.selfSubscriber @ aurelia-templating.js:2937
call @ aurelia-templating.js:2809
flushMicroTaskQueue @ aurelia-task-queue.js:110
(anonymous function) @ aurelia-task-queue.js:60
```

jdanyow:
@DarienFord can you confirm you have templating-resources 0.16.1?
https://github.com/aurelia/templating-resources/releases/tag/0.16.1

You should be able to execute `jspm update` and get all the bug fixes from today.  There's one more coming in binding, unrelated to this issue though

DarienFord:
Confirmed, 0.16.1 does not fix this.

martingust:
We are doing some big refactoring of the repeater to make it pluggable. Plan is to get that and fix for this in by the weekend.

martingust:
Fixed in https://github.com/aurelia/templating-resources/commit/1d96e6eea42760fa78580fce4f6b26fa57ed8ba3

PR #124, opened 2015-10-19
re-bind children on repeat re-bind

### LINKED ISSUE #121: repeat.for bound to an array does not bind correctly to an object upon re-bind within if.bind
Similar to #120, in this scenario assume you have a list of objects which you want to choose between to see details.  The details view is within an ```if.bind```.
The first time ```if.bind``` evals to true, the ```repeat.for``` works.  The subsequent times, it does not evaluate the string interpolations.  

e.g.
```html
<div if.bind="selectedItem">
  <ul >
    <li  repeat.for="sample of templateSamples">${sample.TemplateName}</li> 
  </ul>
</div>
```

martingust:
This is not ready to be merged yet. Will need to handle the Map scenario as well (#120).

@EisenbergEffect @jdanyow Please see if this makes sense to you guys or if there is a better way to handle this. Basically I null out all model refs on `unbind` [here](https://github.com/aurelia/templating-resources/blob/982a328ddd657874975b2347dd2173f314648d9f/src/repeat.js#L97) and on repeat re-bind I re-create the models from the last bound items [here](https://github.com/aurelia/templating-resources/blob/982a328ddd657874975b2347dd2173f314648d9f/src/repeat.js#L72).

Fixes #121

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -64,8 +64,14 @@ export class Repeat {
         this.callContext = 'handleMapChangeRecords';
         this.collectionObserver.subscribe(this.callContext, this);
       } else if (items instanceof Array) {
-        let splices = calcSplices(items, 0, items.length, this.lastBoundItems, 0, this.lastBoundItems.length);
+        let lastBoundItems = this.lastBoundItems;
+        let splices = calcSplices(items, 0, items.length, lastBoundItems, 0, lastBoundItems.length);
         this.collectionObserver = this.observerLocator.getArrayObserver(items);
+
+        let children = viewSlot.children;
+        for (let i = 0, ii = children.length; i < ii; ++i) {
+          children[i].bindingContext = this.createFullBindingContext(lastBoundItems[i], i, lastBoundItems.length);
+        }
 
         this.handleSplices(items, splices);
         this.lastBoundItems = this.oldItems = null;
@@ -86,6 +92,11 @@ export class Repeat {
 
     if (this.items instanceof Array) {
       this.lastBoundItems = this.items.slice(0);
+    }
+
+    let children = this.viewSlot.children;
+    for(let i = 0, ii = children.length; i < ii; ++i) {
+      children[i].bindingContext = null;
     }
 
     this.unsubscribeCollection();
@@ -296,7 +307,7 @@ export class Repeat {
 
   handleAddedSplices(array, splices) {
     let spliceIndex;
-    let spliceIndexLow;
+    let spliceIndexLow = 0;
     let arrayLength = array.length;
     for (let i = 0, ii = splices.length; i < ii; ++i) {
       let splice = splices[i];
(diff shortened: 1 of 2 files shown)


jdanyow:
@martingust I think we may need @EisenbergEffect to confirm this but I believe when the repeat unbinds it should be releasing all references to the model.  
I think this would include:
- nulling out properties on the repeat instance like items, oldItems and lastBoundItems
- calling `unbind` on all children (vs doing [this](https://github.com/aurelia/templating-resources/blob/982a328ddd657874975b2347dd2173f314648d9f/src/repeat.js#L99))
- removing the children from the view slot (maybe this takes care of calling `unbind` on the views, not sure)
- unsubscribing from the array/map/set changes

It's my understanding that `unbind` is an aurelia component's chance to clean everything up to avoid memory leaks and should be essentially the exact opposite of what happens in the `bind` method. 

Another thing to check is when the collection mutates or property-changes and causes children to be removed from the view-slot, are the children `unbind`ed?  The view-slot might be taking care of this, not sure.

Definitely need @EisenbergEffect to confirm the above ^^^ - I don't want to be sending you down the wrong path.

EisenbergEffect:
@martingust @jdanyow This is a tricky scenario. Take the case where the repeater is inside an `if`. That `if` triggers an unbind on it's view when it is false, However, a few seconds later it could trigger with the opposite value and be displayed again. I had originally this scenario in mind when I kept the `oldItems` around, in order to make the toggling in a repeater more efficient. However, I think this was probably a mistake and likely resulted in memory leaks and general inconsistency.

So, I think we need to follow what @jdanyow is saying here. Treat unbind as an opposite of bind so that it completely tears down the old views, making sure to enforce the lifecycle. If you return the view to the cache, the factory will take care of making sure unbind gets called.

Also, as part of this work, I'm wondering if we need to do some internal refactoring to introduce a strategy pattern. Have different strategy implementations for Array, Map, Set, etc. which each implement the appropriate logic for those scenarios. Then whenever the collection is set, we can select the appropriate strategy. This would make the code a bit easier to follow I think.

What do you guys think?

martingust:
@EisenbergEffect @jdanyow Great, makes sense to treat `unbind` opposite to `bind` despite the performance deterioration which could be pretty significant in the toggle scenario on big lists, even when caching views. Will start working on that implementation. 

Strategy pattern seems appropriate here, will create an issue for that.

Thanks for great feedback!

EisenbergEffect:
@martingust Should I close this PR and wait for your more extensive repeat fix/refactor PR instead?

martingust:
@EisenbergEffect Yes, let's close this. Will get this in along with the refactoring.

DarienFord:
@martingust Any idea on a timeframe for this?
It's blocking my application pretty heavily.

martingust:
@DarienFord Doing some big refactoring of the repeater right now, that and the fix should be in by this weekend. You can follow the work here https://github.com/aurelia/templating-resources/pull/130. Hang in there.

PR #126, opened 2015-10-22
feat(binding-behaviors): add throttle, debounce, oneTime, oneWay and …

jdanyow:
…twoWay behaviors

### DIFF
--- a/src/debounce-binding-behavior.js
+++ b/src/debounce-binding-behavior.js
@@ -1,0 +1,51 @@
+import {bindingMode} from 'aurelia-binding';
+
+function debounce(newValue) {
+  let state = this.debounceState;
+  if (state.immediate) {
+    state.immediate = false;
+    this.debouncedMethod(newValue);
+    return;
+  }
+  clearTimeout(state.timeoutId);
+  state.timeoutId = setTimeout(
+    () => this.debouncedMethod(newValue),
+    state.delay);
+}
+
+export class DebounceBindingBehavior {
+  bind(binding, source, delay = 200) {
+    // determine which method to debounce.
+    let methodToDebounce = 'updateTarget'; // one-way bindings or interpolation bindings
+    if (binding.callSource) {
+      methodToDebounce = 'callSource';     // listener and call bindings
+    } else if (binding.updateSource && binding.mode === bindingMode.twoWay) {
+      methodToDebounce = 'updateSource';   // two-way bindings
+    }
+
+    // stash the original method and it's name.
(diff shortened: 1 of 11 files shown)


EisenbergEffect:
@jdanyow Can you also create a `signal` binding behavior?

jdanyow:
sure

jdanyow:
# Signal Binding Behavior

@EisenbergEffect I was researching what the `signal` behavior needed to do based on what we've mentioned in various github issues:

## Use Cases
1. translation:  ability to signal binding to update the target due to the culture/locale changing (https://github.com/aurelia/i18n/issues/6#issuecomment-137452771 and https://github.com/aurelia/framework/issues/144#issuecomment-115430168)

2. "fuzzy time" eg a binding that results in `jdanyow commented an hour ago`.  Need the ability to periodically signal the binding to refresh (https://github.com/aurelia/templating-binding/issues/25#issuecomment-87060863).

3. general purpose "my source expression isn't fully observable but isn't dirty-checked either" (https://github.com/aurelia/framework/issues/32)

## Requirements/Scope
- The `signal` binding behavior should give developers the ability to "signal" a binding or collection of bindings by "name".
- Signaling a binding should update the target (the view) with the source's (view-model's) current value.
- the `signal` binding behavior is only applicable to Bindings and InterpolationBindings.  It is not applicable to Call/Listener/Name type bindings, for these cases an error will be thrown.
- the `signal` binding behavior is not applicable to one-time bindings.  Currently there is no reason to not support one-time bindings but I'm afraid supporting one-time bindings will prevent us from further optimizing one-time binding scenarios. 

---

I'm implementing this behavior as-described above- let me know if any of this needs to be changed.

---
https://github.com/aurelia/templating-binding/issues/25

EisenbergEffect:
This sounds great!

src/debounce-binding-behavior.js
```diff
@@ -0,0 +1,47 @@
...
+      timeoutId: null
+    };
+  }
+
+  unbind(binding, source) {
+    // restore the state of the binding.
+    let methodToRestore = binding.debouncedMethod.originalName;
+    binding[methodToRestore] = binding.debouncedMethod;
+    binding.debouncedMethod = null;
+    if (binding.debounceState.timeoutId !== null) {
```
bryanrsmith: `clearTimeout` ignores invalid timeout IDs, so you don't need to guard this.

jdanyow:
Binding behavior info:  http://stackoverflow.com/a/33490150/725866

PR #128, opened 2015-10-26
feat(repeat): scope inheritance

jdanyow:
I think we're favoring this approach over https://github.com/aurelia/binding/pull/201

cc @EisenbergEffect @bryanrsmith @PWKad


EisenbergEffect:
Closing this since I think we'll be handling this another way.

PR #129, opened 2015-10-26
chore(with): remove deprecated 'with' command

jdanyow:


EisenbergEffect:
I'm going to close this since we think we can fix it now ;)

PR #130, opened 2015-10-27
Repeat strategy pattern

martingust:
@EisenbergEffect This is work in progress to (possible) get @jdanyow started doing his work and for you to review. The implementation should be pretty ready, but needs more testing, might want to break the strategy classes out to different files. Unit tests need more work and probably not green atm.

### DIFF
--- a/src/array-collection-strategy.js
+++ b/src/array-collection-strategy.js
@@ -1,0 +1,73 @@
+import {CollectionStrategy} from './collection-strategy';
+
+export class ArrayCollectionStrategy extends CollectionStrategy {
+  processItems(items) {
+    let i;
+    let ii;
+    let row;
+    let view;
+    let observer;
+    this.items = items;
+    for (i = 0, ii = items.length; i < ii; ++i) {
+      row = super.createFullBindingContext(items[i], i, ii);
+      view = this.viewFactory.create(row);
+      this.viewSlot.add(view);
+    }
+  }
+
+  getCollectionObserver(items){
+    return this.observerLocator.getArrayObserver(items);
+  }
+
+  handleChanges(array, splices) {
+    let removeDelta = 0;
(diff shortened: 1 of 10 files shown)


martingust:
Maybe the strategy locator should use a Map lookup table [here](https://github.com/aurelia/templating-resources/blob/strategy-pattern/src/array-collection-strategy.js#L12-L18) for user to overwrite and add additional strategies.

EisenbergEffect:
Yes, we should definitely do something like that. I'm wondering if the container could handle it directly somehow, maybe with a custom resolver. I need to think about it. Do the strategies have state? I'm also wondering if some of this should be pushed into binding. @jdanyow We had some good  feedback from jods about making collection observation easier, in general. Maybe these strategies are the start of the work for that. I haven't reviewed yet. Need to hope on a plane. Will take a look later today.

martingust:
@EisenbergEffect Yes, they have state. I synchronised some more work. Strategies split into different files and fixed up the unit tests. Though not able to get `createModelForUnitTest` to work after getting latest from a few other repos. Probably need to update more repos.

EisenbergEffect:
@martingust How are we on this. Ready to merge?

martingust:
@EisenbergEffect Not yet. I'm going to do some more work/testing in a few hours on this and hopefully wrap it up. Does the implementation look alright to you?

EisenbergEffect:
@martingust OOPS! I just merged it.

EisenbergEffect:
I also updated it to work with the latest master of templating...

stoffeastrom:
haha :) trigger happy today ;P

EisenbergEffect:
If you make additional changes, can you submit another PR with the new stuff? Sorry, I was a bit hasty...

martingust:
haha oops! Yeah, I'll create new PR for additional work.

PR #134, opened 2015-11-02
feat(repeat): observe inner expression

jdanyow:
Fixes aurelia/binding#183

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -1,7 +1,20 @@
 /*eslint no-loop-func:0, no-unused-vars:0*/
 import {inject} from 'aurelia-dependency-injection';
-import {ObserverLocator, calcSplices, getChangeRecords} from 'aurelia-binding';
-import {BoundViewFactory, ViewSlot, customAttribute, bindable, templateController} from 'aurelia-templating';
+import {
+  ObserverLocator,
+  getChangeRecords,
+  BindingBehavior,
+  ValueConverter
+} from 'aurelia-binding';
+import {
+  BoundViewFactory,
+  TargetInstruction,
+  ViewSlot,
+  ViewResources,
+  customAttribute,
+  bindable,
+  templateController
+} from 'aurelia-templating';
 import {CollectionStrategyLocator} from './collection-strategy-locator';
 
 /**
@@ -10,12 +23,14 @@ import {CollectionStrategyLocator} from './collection-strategy-locator';
 * @class Repeat
 * @constructor
 * @param {BoundViewFactory} viewFactory The factory generating the view
+* @param {TargetInstruction} instruction The view instruction
 * @param {ViewSlot} viewSlot The slot the view is injected in to
+* @param {ViewResources} viewResources The view resources
 * @param {ObserverLocator} observerLocator The observer locator instance
 */
 @customAttribute('repeat')
 @templateController
-@inject(BoundViewFactory, ViewSlot, ObserverLocator, CollectionStrategyLocator)
+@inject(BoundViewFactory, TargetInstruction, ViewSlot, ViewResources, ObserverLocator, CollectionStrategyLocator)
 export class Repeat {
   /**
   * List of items to bind the repeater to
@@ -27,9 +42,11 @@ export class Repeat {
   @bindable local
   @bindable key
   @bindable value
-  constructor(viewFactory, viewSlot, observerLocator, collectionStrategyLocator) {
+  constructor(viewFactory, instruction, viewSlot, viewResources, observerLocator, collectionStrategyLocator) {
     this.viewFactory = viewFactory;
+    this.instruction = instruction;
     this.viewSlot = viewSlot;
+    this.lookupFunctions = viewResources.lookupFunctions;
     this.observerLocator = observerLocator;
     this.local = 'item';
     this.key = 'key';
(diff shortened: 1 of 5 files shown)


jdanyow:
@EisenbergEffect what do you think about doing something along these lines to resolve https://github.com/aurelia/binding/issues/183 ?

EisenbergEffect:
I say, give it a go and see how it turns out when fully integrated.

EisenbergEffect:
Was this meant to be merged?

jdanyow:
it's not ready yet- I need to refactor it to work with the strategy stuff, add tests, etc

src/repeat.js
```diff
@@ -94,17 +115,74 @@ export class Repeat {
...
+  let unwrapped = false;
+  while (expression instanceof BindingBehavior) {
+    expression = expression.expression;
+  }
+  while (expression instanceof ValueConverter) {
+    expression = expression.expression;
+    unwrapped = true;
+  }
+  return unwrapped ? expression : null;
 }
```
jdanyow: @EisenbergEffect  / @martingust 
Should the `getSourceExpression` and `unwrapExpression` functions go in a separate file?  If so, what should we name it? `expression-utilities.js`?

src/repeat.js
```diff
@@ -94,17 +115,74 @@ export class Repeat {
...
   }
+
+  handleInnerCollectionChanges(collection, changes) {
+    let newItems = this.sourceExpression.evaluate(this.scope, this.lookupFunctions);
+    if (newItems === this.items) {
+      return;
+    }
+    this.items = newItems;
+    this.itemsChanged();
+  }
```
jdanyow: @martingust does this look right?  It works but I'm wondering if there's a more efficient way.  Some background:

When the `handleInnerCollectionChanges` method is called it means the "inner collection" (the collection that is being converted by a value converter(s)) has mutated.  Previously this "inner collection" wasn't observed.  Line 161 gets the "converted" / "actual" collection value and line 165 assigns it to the repeat's `items` property so we can essentially treat this as-if the `items` property changed.

src/repeat.js
```diff
@@ -94,17 +115,74 @@ export class Repeat {
...
+      this.collectionObserver = this.observerLocator.getArrayObserver(items);
+    } else if (items instanceof Map) {
+      this.collectionObserver = this.observerLocator.getMapObserver(items);
+    } else {
+      return false;
+    }
+    this.callContext = 'handleInnerCollectionChanges';
+    this.collectionObserver.subscribe(this.callContext, this);
+    return true;
+  }
```
jdanyow: Could this be integrated into the strategy logic somehow?

martingust: Yes, I think that would be nice. I wonder if we could only have the `observeCollection` method and set the items conditionally based on what `getInnerCollection` returns? In order to handle all collection type specifics in the strategies. Perhaps something like this (untested): 

```javascript
observeCollection() {
    let innerCollection = this.getInnerCollection();
    if (innerCollection) {
      items = innerCollection; 
      this.callContext = 'handleInnerCollectionChanges';
    } else {
      items = this.items;
      this.callContext = 'handleCollectionChanges';
    }       
    
    this.collectionObserver = this.collectionStrategy.getCollectionObserver(items);
    if (this.collectionObserver) {
      this.collectionObserver.subscribe(this.callContext, this);
    }
  }
```

src/repeat.js
martingust: Looks good to me.

EisenbergEffect:
Not sure if we can get this in today. But, if you guys can both continue to work it out by Friday, it would be nice to have for the beta.

martingust:
@EisenbergEffect / @jdanyow I have looked through, it looks good to me. Just some minor refactoring feedback.

src/repeat.js
martingust: I think it's fine as is, only repeat using those functions.

EisenbergEffect:
Is it ready enough to merge now? @martingust @jdanyow

martingust:
@jdanyow Would have to decide that. Looks good to me, but I haven't tested it.

src/repeat.js
jdanyow: That makes sense- the one thing we'd need to account for is the "inner collection" could be of a different type than the collection returned by the value converter.  

Contrived example: `repeat.for="myMap| toArray | sort:'key' | take:5"`.  The inner collection is a Map instance and the outer collection/strategy is an array.

EisenbergEffect:
@jdanyow Just waiting on the final word from you. Let me know and I'll put it in. If not we can let it bake for a few days and try to get it into the beta.

jdanyow:
Let's put it in.  There's [one piece](https://github.com/aurelia/templating-resources/pull/134#discussion_r44225396) the needs some refactoring to fit into the repeat's strategy pattern properly but I think it would be good to get it in ahead of the beta so that it sees some use in case there's a defect in some edge case that's been introduced.  I can work on the refactoring this week.

does that sound reasonable?

jdanyow:
by the way, here's the integration test I've been using:

```html
<template>
  <form submit.delegate="items.push(a.value)" style="display: block">
    <label>add: <input ref="a"></label>
    <button type="submit">Add</button>
  </form>
  <label style="display: block">filter: <input ref="f"></label>
  <div repeat.for="item of items | filter:f.value | take:20 & debounce:2000">${item}</div>
</template>
```

```js
export class App {
  items = [];
}
```

src/repeat.js
martingust: I see. Maybe we could move out the getCollectionObserver from the strategies, and based on collection type return corret observer or move that to binding? What you think?

jdanyow: What if we added `getObserver(observerLocator, collection)` to the `CollectionStrategyLocator`?  It could use the `matchers` to find the appropriate strategy class.  Then we'd need a static method on the strategy classes that returns an observer when given the observerLocator and collection.  Is it weird to have a static method be part of the strategy interface?

EisenbergEffect: This has not come up with respect to a repeater bug. Hopefully, we will have a repro soon. For reference, here's the discussion link: [November 10, 2015 10:13 PM](https://gitter.im/Aurelia/Discuss?at=5642b25bfa021fde389d03de)

PR #146, opened 2015-11-11
fix(repeat): handle collection instance changes in repeat

### LINKED ISSUE #144: handle collection instance changes in repeat
```html
<template>
  <div repeat.for="foo of bar">${foo}</div>
</template>
```
```javascript
export class App {
  attached() {
    // error will occur when this executes:
    this.bar = ['a', 'b', 'c'];

    // not sure what will happen when this executes:
    setTimeout(() =>  this.bar = new Map([['a', 'b'], ['c', 'd']]), 5000);

    // not sure what will happen when this executes:
    setTimeout(() =>  this.bar = 10, 10000);

    // not sure what will happen when this executes:
    setTimeout(() =>  this.bar = null, 15000);   
  }
}
```

cc @diegograsso @martingust @EisenbergEffect

martingust:
Fixes #144

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -60,21 +60,20 @@ export class Repeat {
 
   bind(bindingContext, overrideContext) {
     let items = this.items;
+    this.sourceExpression = getSourceExpression(this.instruction, 'repeat.for');
+    this.scope = { bindingContext, overrideContext };
     if (items === undefined || items === null) {
       return;
     }
-
-    this.sourceExpression = getSourceExpression(this.instruction, 'repeat.for');
-    this.scope = { bindingContext, overrideContext };
-    this.collectionStrategy = this.collectionStrategyLocator.getStrategy(this.items);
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Great. Will get this into a release today.

jdanyow:
@martingust I pulled this PR down locally and it's working perfectly with the integration tests that I have.  One of the unit tests is failing when I run the test suite- do the testsl need syncing up with the refactoring?

The only other comment/question I had was around the `if(this.scope)` statement [here](https://github.com/aurelia/templating-resources/pull/146/files#diff-3a07a9d53005f671f25879cd1321a988R110).  Is that required?  I was thinking it meant that processItems was being called after `unbind`.  If we land in processItems and `this.scope` is null it's probably better to `return` because it means the repeat is "unbound".

EisenbergEffect:
Merged. Not sure why GH didn't pick it up.

PR #153, opened 2015-11-17
test(repeat): integration tests

jdanyow:
This is ready to merge if it looks good.


jdanyow:
todo: add test for [this](https://github.com/aurelia/binding/issues/233) use case

PR #154, opened 2015-11-18
fix(repeat): obey one-time binding behavior

jdanyow:


EisenbergEffect:
Love seeing these scenarios represented in the tests! Great work.

PR #168, opened 2015-12-17
tests won't run after `git clone ... ; npm install ; jspm install ; gulp test`

ohjames:
Without the added `.bind` jasmine thinks that the parameter `viewsRequireLifecycle` is the `done` argument, the current version pulled by the package.json dependency issues a warning for this as `describe` functions do not take a `done` parameter.


EisenbergEffect:
Thanks for the contribution! Sorry for the delay in getting it merged...busy time of year right now.

Issue #167, opened 2015-12-17
repeat.for: loses sync between array and view when array updates happen during leave animation


ohjames:
When events are removed from the view, they remain in viewSlot.children until the animation completes, see https://github.com/aurelia/templating/blob/master/src/view-slot.js#L204

The `removing` state is accounted for here during the mutation: https://github.com/aurelia/templating-resources/blob/master/src/array-repeat-strategy.js#L115 (by taking into account `rmPromises.length`, this works as the splices are ordered).

Say this animation takes 1 second, and 200ms later an insert happens at the end of the array. At this stage `viewSlot.insert` is called with the index of the item in the array, yet the index within the `viewSlot` is different due to the animations not having completed (`viewSlot.children` now has a different number of elements to the array). Now `rmPromises.length` can't be used to factor in the indexes of the removed items from a previous set of splices and the indexes are all wrong.

This leads to the view going badly out of sync with the array, this is particularly bad if you are using say `splice(idx)`. This can end up splicing elements off of the end that should not be removed, then a later update can supply an index that should exist within `viewSlot.children` but doesn't, throwing the exception `Cannot read property 'firstChild' of undefined`.

ohjames:
Calling @EisenbergEffect @martingust @jdanyow this is a serious issue with `repeat.for` affecting all `animator` libraries. It will affect any user where arrays update during animation leading to extremely bad results.

ohjames:
The `+ rmPromises.length` correction factor also bothers me. I can see how it works to fix the offset of splices at a higher index, but what if the first splice is at a higher index than the second splice... in that case wouldn't the `+1` incorrectly be applied to the second splice (at the lower index)? It seems to me it should only affect splices at a higher index, (and after being corrected for intermediate additive spices). Would really appreciate some feedback on that point.

EisenbergEffect:
@martingust @jdanyow Can you guys work s with @ohjames to see if we can resolve that. We may also need help from @zewa666 if this involves the css animator. You all know more about this code than I do at this point...

martingust:
@ohjames I think you are right, the `rmPromises` should only affect splices at a higher index. Do you have an example project or plunker that shows the issue you described, that would help a lot to fully see the problem. 

However when trying this out I think I have found a related (or same) bug. 

Let's say I have an array of 4 items: `array = ['foo', 'bar', 'baz', 'qux'];` bound to a `repeat`.

[...]

PR #169, opened 2015-12-18
fix updates during leave animation

ohjames:
This is an in process PR for fixing #167 

It currently just includes a test that demonstrates the issue.

### DIFF
--- a/src/array-repeat-strategy.js
+++ b/src/array-repeat-strategy.js
@@ -1,4 +1,5 @@
 import {createFullOverrideContext, updateOverrideContexts, updateOneTimeBinding} from './repeat-utilities';
+import {mergeSplice} from 'aurelia-binding';
 
 /**
 * A strategy for repeating a template over an array.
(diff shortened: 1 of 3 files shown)


ohjames:
@martingust have a fix here. It works almost perfectly but if we can coalesce the splices (see https://github.com/ohjames/templating-resources/blob/0549456ee2c5645e78cc1297ed2dfa30ab68abe4/src/array-repeat-strategy.js#L121) then it will be much better/simpler and ready for commit/review.

The test passes anyway with the fix as it is.

ohjames:
Okay I have the normalising in place now in e557cfb should be ready to merge. With this fix things run a lot better on my website, no more exceptions thrown from aurelia :) It'll need a small change to `aurelia-binding`, to export `mergeSplice`.

On the down side.. they still aren't quite perfect. Although I think the only remaining bug is duplicate elements appear in the view (over the top of non-duplicated elements that should be in their place).

martingust:
@ohjames Looks solid, unit tests passing and I ran a few succesful tests scenarios in skeleton. The remaining bug is that something you think you can get in a fix with this PR or create a new issue for that?

ohjames:
@martingust Cool, I'll open a new github issue for the remaining bug. Would be great to get this merged, it improves things *a lot*.

EisenbergEffect:
@ohjames I can merge and do a release on Monday assuming this is an incremental fix and doesn't require your next PR. I can release that when it's available too...I just want to make sure I don't need to wait in order to sync things without breaking people. Does that make sense?

ohjames:
@EisenbergEffect It would be great to release this fix ASAP I think given how severe the issue is and how easy it is to manifest, it depends on the PR against aurelia/binding only.

EisenbergEffect:
@ohjames  Just to be clear, in order to release this, I must first release this: https://github.com/aurelia/binding/pull/265 Correct?

ohjames:
@EisenbergEffect affirmative

PR #195, opened 2016-03-03
fix(array-repeat-strategy): prevent array being manipulated during an…

### LINKED ISSUE #183: repeat.for + animations still not production ready, view can desync with array leading to duplicates and loss of order.
After my fixes in the animator (https://github.com/aurelia/animator-css/issues/35 and https://github.com/aurelia/animator-css/pull/33) and my fix in this project (#167) I was hoping to see the last of issues this critical but the combination of `repeat.for` and `au-animate` still leads to duplicate entries in the view, the view entries not being sorted correctly and items missing from the end of the repetition.

I've noticed this issue many times over the last few weeks, I can't replicate it reliably in a test though, must be the result of a race condition.

If I add the elements `['a'..'z']` using only splices (but say, adding 'a', 'b', 'z'; then splicing in 'c', 'd', 'e' and so on until I have all 26 elements) rather than seeing 26 elements in order from 'a' to 'z' I *will* see 26 elements. However I end up with the order like `['a', 'b', 'c', 'a', 'b', 'd', 'e', 'f', 'e' ... 'v']` (and elements after 'v' will be missing, the faulty duplicates "push them off the end") 

I see the issue when doing only inserts after about 10-20% of updates, so I started throttling updates so they only occur every 1 second, this makes the issue occur less frequently but it still happens enough to be annoying. It seems to happen in mobile chrome more than desktop chrome/firefox, perhaps the race condition shows up more frequently on machines with less CPUs.

You can see the issue on this website also: http://kchomp.co you'll need to add a few subscriptions for it to trigger.

I really would advise in the documentation for aurelia that animations + repeat.for are not ready yet also, I think this bug will take a while to track down and until then I wouldn't want other people relying on basic functionality that is critically broken.

martingust:
…imation

This will make a copy of the array to prevent it being manipulated by
another set of changes during animation.

Fixes #183

### DIFF
--- a/src/array-repeat-strategy.js
+++ b/src/array-repeat-strategy.js
@@ -113,7 +113,7 @@ export class ArrayRepeatStrategy {
       return;
     }
 
-    let maybePromise = this._runSplices(repeat, array, splices);
+    let maybePromise = this._runSplices(repeat, array.slice(0), splices);
     if (maybePromise instanceof Promise) {
       let queuedSplices = repeat.__queuedSplices = [];
 


EisenbergEffect:
Thanks for working on this @martingust 

Before we release this, I'd like to know that this isn't going to cause us any performance or memory pressure issues. @jdanyow Can you take a look at this? Can we run some benchmarks and take a look at how this affects dbmonster as well? I'd hate to do something that set us back in a big way. If this is only needed when animation is involved, maybe we can only clone the array in those circumstances?

What do you all think?

martingust:
@EisenbergEffect Good points. We should probably only clone the array [if animation](https://github.com/aurelia/templating-resources/blob/master/src/analyze-view-factory.js#L45) is involved. I don't think this code will get hit in the dbmonster though. @jdanyow Thoughts?

martingust:
Just ran some very basic heap profiles and timeline in chrome dev tool, adding 1 item every 1ms till 1000 items, could not see any noticeable difference between cloning the array or not. Not sure if I tested the right thing here though. It's going to be interesting to see @jdanyow take on this or some better benchmarks.

EisenbergEffect:
That's a good sign. Let's see if @jdanyow has anything to add.

jdanyow:
This looks good to me.  If you really wanted to avoid the slice you could do something like this:

```js
if (repeat.viewsRequireLifecycle) {
  ... slice it up ...
}
```

`viewsRequireLifecycle` will be true when the template uses animation:  https://github.com/aurelia/templating-resources/blob/master/src/analyze-view-factory.js#L45

EisenbergEffect:
@jdanyow Nice!
@martingust Can we make that change? If so, I can get it merged and released tonight.

Thanks guys!

EisenbergEffect:
As it turns out, the lifecycle check is in another method so when this method is called we already are sure that we need the lifecycle. I just left it as is :) Releasing shortly.

Good job @martingust This was a big one to squash!

PR #196, opened 2016-03-03
Silently ignore invalid collections

gingerik:
First of all, I am not sure whether this PR is _the_ desired solution to my issue, but at least it would have made things less confusing to me.

I accidentally set a string rather than an array as replacement for an existing bound collection and got the following error: `cannot call method 'getCollectionObserver' of null`. Diving into the stack trace showed me the error was in the method `_observeCollection()`. I had to step trough this package to find out what actually caused this error and found out no repeat strategy could be selected (which makes sense). I am not sure what the desired behavior of Aurelia would be in such a case, but a general JavaScript error seems to be unwanted.

Please let me know if any additional information is required, I do not have the original code at hand right now.

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -120,6 +120,9 @@ export class Repeat {
 
     let items = this.items;
     this.strategy = this.strategyLocator.getStrategy(items);
+    if (!this.strategy) {
+      throw new Error(`Value for '${this.sourceExpression}' is non-repeatable`);
+    }
 
     if (!this.isOneTime && !this._observeInnerCollection()) {
       this._observeCollection();


EisenbergEffect:
@martingust Can you review this? I do think we should so something. I'm not sure if we should silently ignore though. Perhaps we can take this PR, but amend it with some logging that explains what type was received and how it can't be observed as a collection, and what to do about it. Basically, a more meaningful message, but  without crashing the app. What do you think?

atsu85:
> I'm not sure if we should silently ignore though

I think silently ignoring the error caused by wrong type is terrible idea. When error is not produced, the problem may remain unnoticed and cause painful debugging sessions afterwards (even if just a warning is logged before). My suggestion is failing fast whenever possible, with meaningful error message, so that the framework would support creating solid applications.

EisenbergEffect:
I agree, a silent error isn't desired. However, I'm concerned that throwing an error could break some valid scenarios as well. I'd like to get some feedback from @martingust and @jdanyow

src/repeat.js
```diff
@@ -120,6 +120,9 @@ export class Repeat {
 
     let items = this.items;
     this.strategy = this.strategyLocator.getStrategy(items);
+    if (!this.strategy) {
+      return;
```
jdanyow: I think throwing instead of returning would be good here. It would keep the behavior the same while improving the developer experience by giving clear information why things are broken.  I think you could even include `this.sourceExpression.toString()` in the error message so the developer will know which binding expression is returning a non-repeatable value.

EisenbergEffect: That sounds reasonable to me. Let's do it. @gingerik Would you be interested in amending your PR to throw an error  based on @jdanyow 's recommendations? If you aren't comfortable with that we could take this PR and then make the change ourselves. Let me know.

gingerik:
@EisenbergEffect @jdanyow -- I've updated the PR so an error is thrown instead.

martingust:
@gingerik Nice!

Sorry for being late to the conversation but I agree with you guys, if not strategy is found it should throw error, but with a better description just like @gingerik provided.

EisenbergEffect:
Cool. I can merge this and release if @martingust or @jdanyow give me an official approval.

jdanyow:
looks good- thanks @gingerik :dancers:

PR #197, opened 2016-03-04
perf(array-repeat-strategy): call array.slice with all arguments

Gheoan:
Also, is `array.slice` the fastest way to clone an array?

### DIFF
--- a/src/array-repeat-strategy.js
+++ b/src/array-repeat-strategy.js
@@ -109,11 +109,11 @@ export class ArrayRepeatStrategy {
         let {index, removed, addedCount} = splices[i];
         mergeSplice(repeat.__queuedSplices, index, removed, addedCount);
       }
-      repeat.__array = array.slice(0);
+      repeat.__array = array.slice(0, array.length);
       return;
     }
 
-    let maybePromise = this._runSplices(repeat, array.slice(0), splices);
+    let maybePromise = this._runSplices(repeat, array.slice(0, array.length), splices);
     if (maybePromise instanceof Promise) {
       let queuedSplices = repeat.__queuedSplices = [];
 


EisenbergEffect:
@martingust @jdanyow What do you guys think?

src/array-repeat-strategy.js
```diff
@@ -109,11 +109,11 @@ export class ArrayRepeatStrategy {
         let {index, removed, addedCount} = splices[i];
         mergeSplice(repeat.__queuedSplices, index, removed, addedCount);
       }
-      repeat.__array = array.slice(0);
+      repeat.__array = array.slice(0, array.length);
```
alvarezmario: @Gheoan it seems is the best option, and also, 0 and array.lenght are the default values for slice, so you can safely remove them

alvarezmario: Well, in fact, according to this http://jsperf.com/new-array-vs-splice-vs-slice/113, `slice` is not the fastest, is at the medium of the table. Using a while loop seems to be faster, but is really worth?

Gheoan:
Looking at the [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.slice) again, I was wrong :sweat_smile:, passing the 2nd argument causes an extra call to `ToInteger` so omitting it should be faster. 
> If _end_ is **undefined**, let _relativeEnd_ be _len_; else let _relativeEnd_ be ToInteger(end).

EisenbergEffect:
It's probably very browser specific. I just tested FF on Mac and slice was faster than the while loop, but only by a little bit.

EisenbergEffect:
I think we can leave it as is for now.

PR #198, opened 2016-03-04
style(array-repeat-strategy): add comments about array.slice usage

Gheoan:
I had no idea array.slice can be used to clone an Array, so the comments should help others reading the code.


EisenbergEffect:
Sounds good to me :)

PR #201, opened 2016-03-09
introduce AbstractRepeater

mbroadst:
The `AbstractRepeater` is a base class for all attributes or elements that repeat views/templates over a collection. It's meant primarily to provide an abstraction for anyone who wants to create something like the baked in `repeat` attribute. The flexibility of the interface allows you to implement repeat-esque elements over more complicated components (particularly multi-dimensional datasets like a grid, where each "item"/"row" might actually have multiple views).

I think this is a good place to start, there are a number of nitpicks I can point out off the top of my head:
  * `addView` is more like `createAndAddView`
  * `insertView` => `insertViewAt`, as well as the above comment
  * `removeView` => `removeViewAt`
  the list could go on

I'm not sure what kind of conventions y'all prefer, and am not tied whatsoever to the existing naming.

### DIFF
--- a/src/abstract-repeater.js
+++ b/src/abstract-repeater.js
@@ -82,4 +82,14 @@ export class AbstractRepeater {
   removeView(index: number, returnToCache?: boolean, skipAnimation?: boolean) {
     throw new Error('subclass must implement `removeView`');
   }
+
+  /**
+   * Forces a particular view to update it's bindings, called as part of
+   * an in-place processing of items for better performance
+   *
+   * @param {Object} view the target view for bindings updates
+   */
+  updateBindings(view: View) {
+    throw new Error('subclass must implement `updateBindings`');
+  }
 }
--- a/src/abstract-repeater.js
+++ b/src/abstract-repeater.js
@@ -1,0 +1,85 @@
+/**
+* An abstract base class for elements and attributes that repeat
+* views.
+*/
+export class AbstractRepeater {
+  constructor(options) {
+    Object.assign(this, {
+      local: 'items',
+      viewsRequireLifecycle: true
+    }, options);
+  }
+
+  /**
+   * Returns the number of views the repeater knows about.
+   *
+   * @return {Number}  the number of views.
+   */
+  viewCount() {
+    throw new Error('subclass must implement `viewCount`');
+  }
+
+  /**
+   * Returns all of the repeaters views as an array.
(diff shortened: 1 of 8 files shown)


EisenbergEffect:
@jdanyow @martingust I'd love for you guys to review this. It could be a really nice improvement. Maybe there's even some way this could help with virtualization work being done too?

mbroadst:
hey guys, to give a little bit of context about what this opens up you can take a look at the playground for the grid component I've been working on (with invaluable help from @jdanyow, as we descend into the bowels of the templating engine together :smile: )

https://github.com/mbroadst/grid-component/blob/master/src/grid/grid.js

jdanyow:
I think this looks really good. @martingust what do you think? Any changes needed to make this work with the virtualizing repeater?

martingust:
At first look, I think all the methods could be used in `virtual-repeat`, will definitely implement this in UI-Virtualization next. Perhaps even more can be abstracted out along the way. Really nice, great work @mbroadst!

mbroadst:
So is that two +1's, or do you guys want to discuss alternative api naming/other ideas?

EisenbergEffect:
@jdanyow @martingust Shall we merge this?

jdanyow:
@mbroadst - is this still good to merge? Just wanted to confirm because I know you're refactoring some things in your grid based on the `viewRequiresLifecycle` stuff. If so I'll do the perf tests, etc and get it in. I don't expect there to be any issues.

@martingust just to confirm- we're going to merge this in right away and iterate on it as-needed when you get a chance to incorporate this into the virtual repeater.  Does that sound right?

mbroadst:
@jdanyow it's good to merge on my side. If anything, the things I would need for `viewRequiresLifecycle` can be added later (or I think at the moment just exposed). Either way this work needs to get in first

mbroadst:
@jdanyow I've added `updateBindings` to `AbstractRepeater`, that's the final bit of code I think that can be abstracted out of the repeat strategies

martingust:
@jdanyow That sounds right. All non public API:s that we can iterate over along the way. Perf tests would be great if you get a chance too run, nothing I can see that would impact the performance in any direction, but just to be sure.

jdanyow:
tests passing, perf looks good in dbmonster.  Nice job @mbroadst!

PR #202, opened 2016-03-13
fix(if): queue up changes when animating

martingust:
Changes are now being queued when animating, same approach as `repeat`.

@jdanyow @EisenbergEffect If this looks to you guys, this is ready to be merged.

### DIFF
--- a/src/if.js
+++ b/src/if.js
@@ -40,9 +40,34 @@ export class If {
   * @param newValue The new value
   */
   valueChanged(newValue) {
+    if (this.__queuedChanges) {
+      this.__queuedChanges.push(newValue);
+      return;
+    }
+
+    let maybePromise = this._runValueChanged(newValue);
+    if (maybePromise instanceof Promise) {
+      let queuedChanges = this.__queuedChanges = [];
+
+      let runQueuedChanges = () => {
+        if (!queuedChanges.length) {
+          this.__queuedChanges = undefined;
+          return;
+        }
+
+        let nextPromise = this._runValueChanged(queuedChanges.shift()) || Promise.resolve();
+        nextPromise.then(runQueuedChanges);
+      };
+
+      maybePromise.then(runQueuedChanges);
+    }
+  }
+
+  _runValueChanged(newValue) {
     if (!newValue) {
+      let viewOrPromise;
       if (this.view !== null && this.showing) {
-        let viewOrPromise = this.viewSlot.remove(this.view);
+        viewOrPromise = this.viewSlot.remove(this.view);
         if (viewOrPromise instanceof Promise) {
           viewOrPromise.then(() => this.view.unbind());
         } else {
@@ -51,7 +76,7 @@ export class If {
       }
 
       this.showing = false;
-      return;
+      return viewOrPromise;
     }
 
     if (this.view === null) {
@@ -64,7 +89,7 @@ export class If {
 
     if (!this.showing) {
       this.showing = true;
-      this.viewSlot.add(this.view);
+      return this.viewSlot.add(this.view);
     }
   }
 
--- a/test/if.spec.js
+++ b/test/if.spec.js
@@ -104,6 +104,50 @@ describe('if', () => {
 
     expect(view.bind).toHaveBeenCalled();
   });
+  
+  describe('during animation', () => {
+    let delay = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+    let removeWithAnimation = animationDuration => {
+      return function(view) {
+        return delay(animationDuration).then(() => {            
+          this.children = [];
+        });
+      };
+    };
+      
+    let addWithAnimation = animationDuration => {
+      return function(view) {
+        return delay(animationDuration).then(() => {          
+          this.children.push(view);
+        });
+      };
+    };
+
+    beforeEach(() => {   
+      spyOn(viewSlot, 'remove').and.callFake(removeWithAnimation(600));
+      spyOn(viewSlot, 'add').and.callFake(addWithAnimation(0));
+    });
+
+    it('should correctly handle value changes during remove animation', done => {  
+      sut.showing = false;
+      sut.view = new ViewMock();  
+      sut.view.isBound = true;        
+      sut.viewSlot.children = [];
+      sut.valueChanged(true);        
+   
+      delay(200).then(() => {          
+        sut.valueChanged(false);
+        return delay(400);
+      }).then(() => {
+        sut.valueChanged(true);
+        return delay(600);          
+      }).then(() => {  
+        expect(viewSlot.children.length).toEqual(1);
+      })
+      .then(() => done())
+    })
+  });
 });
 
 class ViewSlotMock {
--- a/src/array-repeat-strategy.js
+++ b/src/array-repeat-strategy.js
@@ -106,8 +106,8 @@ export class ArrayRepeatStrategy {
 
       let runQueuedSplices = () => {
         if (!queuedSplices.length) {
-          delete repeat.__queuedSplices;
-          delete repeat.__array;
+          repeat.__queuedSplices = undefined;
+          repeat.__array = undefined;
           return;
         }
 


jdanyow:
looks good to me- super minor feedback- might be better to remove `delete` and use
`if (this.__queuedChanges === undefined) {` for the check and `this.__queuedChanges = undefined;` for the clear.

martingust:
Cool, will change that, in both `if` and `repeat`. Thanks

test/if.spec.js
```diff
@@ -104,11 +104,56 @@ describe('if', () => {
...
+        .then(() => done())
+      })
+    })
+  
 });
 
 class ViewSlotMock {
   remove() {}
-  add () {}
+  add () {}
```
atsu85: i bet this change is unintentional ;)

src/if.js
```diff
@@ -43,20 +43,42 @@ export class If {
...
+    let maybePromise = this._runValueChanged(newValue);
+    if (maybePromise instanceof Promise) {
+      let queuedChanges = this.__queuedChanges = [];
+
+      let runQueuedChanges = () => {
+        if (!queuedChanges.length) {
+          delete this.__queuedChanges;       
+          return;
+        }        
+        let nextPromise = this._runValueChanged(queuedChanges.shift()) || Promise.resolve();
```
atsu85: I suspect, i'm missing smth really obvious, 
but right now it seems to me, that 
1) `let queuedChanges = this.__queuedChanges = [];` always assigns empty array to `queuedChanges`
2) therefore because `queuedChanges.length` is 0, then `!queuedChanges.length` is `true`
3) therefore it seems to me, that **this line is never reached**, as function returns from if block

Is this code written as intended - what am i missing?

martingust: @atsu85 If a new change happens before an animation (ie the method `valueChanged`) has finished the new value will be pushed to `__queuedChanges` on line 47 and `queuedChanges.length` will be `> 0`, does that makes sense?

atsu85: @martingust, thanks for the reply.
> If a new change happens before an animation (ie the method valueChanged) has finished the new value will be pushed to __queuedChanges on line 47

That was clear to me,

however following
> `queuedChanges.length` will be `> 0`

is not clear to me, as it seems to me, that my previous statement:
> 1) `let queuedChanges = this.__queuedChanges = [];` always assigns empty array to `queuedChanges`

... **still holds even if `this.__queuedChanges` is not empty**. Let me create a similar setup (difference is that in this demo `__queuedChanges` is var instead of instance field):
```javascript
var __queuedChanges = [];
__queuedChanges.push({});

var queuedChanges = this.__queuedChanges = [];
console.log("queuedChanges.length =", queuedChanges.length);
```
produces `queuedChanges.length = 0`

martingust: In that case length will be `0`. But if a new change happens before first has finished this `this.__queuedChanges.push(newValue);` will happen after `let queuedChanges = this.__queuedChanges = [];` and before `if (!queuedChanges.length) {`. 

Easiest might be to try this out in an example, you can try out the code in skeleton with [this](https://github.com/aurelia/animator-css/issues/44) example, and add `console.log(queuedChanges.length)` [here](https://github.com/aurelia/templating-resources/blob/if-animation/src/if.js#L60), then trigger a few quick changes you will see a number get logged and decreased by one as each animation has finished until `0` and the function will return from the `if` statement.

atsu85: > I suspect, i'm missing smth really obvious

Yeah, concurrency it is ;) 
> 1) let queuedChanges = this.__queuedChanges = []; always assigns empty array to queuedChanges

This is true, but before callback is executed, newValue is pushed to the empty array and therefore when executing `runQueuedChanges` callback, my second assumption doesn't hold.

Thanks for explaining it and sorry for wasting Your time.

martingust: Yes, that is right. You not wasting my time at all, thanks for asking, makes me happy people care :)

EisenbergEffect:
Yes, definitely NOT a waste of time. With open source, it's really important to distribute the knowledge of how everything is working. These questions are good and help all of us to work together better. Thanks for your involvement @atsu85 !

martingust:
@jdanyow I changed to assign undefined instead of delete the property [here](https://github.com/aurelia/templating-resources/blob/if-animation/src/if.js#L57) also updated the same in `array-repeat-strategy` [here](https://github.com/aurelia/templating-resources/blob/if-animation/src/array-repeat-strategy.js#L124-L125), let me know if that looks better.

jdanyow:
:+1:

EisenbergEffect:
@martingust Can you rebase this on master? I can merge it as soon as that's done. Sorry about that!

EisenbergEffect:
@martingust The conflict comes because @jdanyow removed the micro task queue usage. So, you probably need to just re-implement and test it again with both of these things. Does that make sense?

martingust:
@EisenbergEffect No problem, this PR removes the micro task queue as well so shouldn't be much difference.

src/if.js
```diff
@@ -40,9 +40,34 @@ export class If {
   * @param newValue The new value
   */
   valueChanged(newValue) {
+    if (this.__queuedChanges) {
+      this.__queuedChanges.push(newValue);
```
jods4: @EisenbergEffect @martingust 
I am reviewing the `if` custom attribute in preparation for an `else` PR.
Is a queue really what we want here?
Let's say that there is a 5 sec animation on entry/exit (awefully long) and a button that toggles visibility.
If I spam the button like crazy, say I click 8 times. Do we really expect a 40 sec animation that does in-out-in-out-in-out?

I think a good UX would rather be: when an animation finish, compare with the current expected state. If it's the same -> done. If it's different add/remove appropriately and repeat the check when the animation (if any) ends.

EisenbergEffect: @martingust Can you comment on this?

martingust: @jods4 @EisenbergEffect Agree, that would probably be a better experience.

EisenbergEffect: @jods4 Are you able to submit a PR and @martingust could you review that?

jods4: @EisenbergEffect I'm going to tweak this and add `else` in a PR

EisenbergEffect: Sounds excellent :)

src/if.js
```diff
@@ -40,9 +40,34 @@ export class If {
...
+      maybePromise.then(runQueuedChanges);
+    }
+  }
+
+  _runValueChanged(newValue) {
     if (!newValue) {
+      let viewOrPromise;
       if (this.view !== null && this.showing) {
-        let viewOrPromise = this.viewSlot.remove(this.view);
+        viewOrPromise = this.viewSlot.remove(this.view);
```
jods4: @EisenbergEffect  @martingust  I don't think I'll touch this because it's very edgy, but for the sake of discussion: 
There's a race condition when the `if` is unbound while an exit animation is in progress.
💣 Views can be returned to cache before the animation completes and before they are actually remove from the DOM :(
If the cache is not used, there's a possibility that when the `if` is re-bound, users see the exit animation complete, although that's less bad.
BTW: I'm not very familiar with our animation APIs, is there a way to (optionally?) skip the enter animation on first render?

EisenbergEffect: Currently, there's only a way to skip the animation on remove, but we could add one for add as well in the future.

jods4: From a UX perspective, when I load a page I don't necessarily want to see all elements transition in. 

For example consider a carousel (image slideshow) the first image should probably not slide in from the right when the page is loaded, it's "there". But then I expect each image to slide in when I click on next.

If I understand everything correctly, the `if` will always apply an entry animation, even when first loading the page.

Anyway, that's a story for another day.

What about the race condition?

EisenbergEffect: I think we should address that separately. The solution might be to extend the classes that the animator understands so that we give that control to the developer rather than making assumptions. Let's see if we can fix the if and add the else first and then pick up the other conversation after that.

Issue #189, opened 2016-02-12
Error when requiring a missing css file


thomas-darling:
This error only occurs when we bundle our app - when unbundled, it just continues with no error.

Basically, if the app is bundled and a view `require` a css file that does not exist, the app crashes with a fairly unhelpful error message - and it gets even worse when minified: "b.replace is not a function":

```html
<require from="./app-logo.css"></require>       <-- this file does not exist
```

![error when referencing missing css file](https://cloud.githubusercontent.com/assets/6388663/13007908/81c0d208-d193-11e5-88e4-362e63c669cd.png)

Stepping through the code reveals that the `css` argument, which is normally a string, somehow becomes an empty object if the css file is missing - but again, it only happens when the app is budled:

![breakpoint when referencing missing css file](https://cloud.githubusercontent.com/assets/6388663/13008061/73e8f844-d194-11e5-8f13-997b091d4617.png)

The unhelpful error message aside, it's a really nasty surprise to encounter after making a release, so I would strongly prefer that it outputs a warning or error message both when bundled and unbundled.
Maybe just a warning when unbundled, so it doesn't get in the way of the dev workflow - reason for this is that e.g. the LESS compiler does not create a .css file unless the .less file produces non-empty CSS output.. so even though the .less file exists, it may not actually produce a .css file, either because the .less file is empty, or because it contains only variables or mixins, but not yet any style rules that would produce CSS output.

EisenbergEffect:
We can easily provide a warning message for this. Thanks for letting us know!

PR #205, opened 2016-03-18
fix(css): throw on missing CSS file

gingerik:
See #189. Throws an error on missing CSS files -- both when bundled and unbundled.


EisenbergEffect:
Thanks!

PR #208, opened 2016-03-28
fix(html-resource-plugin): handle query string and mixed casing

jdanyow:


EisenbergEffect:
Nice! Does the css resource require any similar fixups?

jdanyow:
I don't think so- it doesn't look like it computes a name from the load address like the html plugin.

EisenbergEffect:
Ok, cool. Thanks for working on this!

Issue #213, opened 2016-04-17
Performance: Suboptimal repeat element lifecycle


niieani:
* [Example (with `@bindable`)](https://gist.run/?id=47985f620e3632b40f18197e52ecf77f)
* [Example (with `<compose>`)](https://gist.run/?id=bd1ee888a9311027156b29b7df6ac27e)

1. Open either of the examples
2. Open Developer Tools console to see when elements bind/unbind. 
3. Click the checkbox to filter the object with the value '8'.

What happens is that when one toggles the value filter, Aurelia doesn't diff between the unfiltered and filtered elements, but removes **all** of them from the view and recreates all of them again, each time. In the developer console you can see that upon enabling, each of the Components is unbound and bound again, only this time without the filtered value 8. The same thing happens when you toggle it back. This leads to very bad performance when there are more elements - try changing the number of elements from 10 to 1000 in the for loop of `app`. With 1000 elements my MBP takes at least 1-2 seconds to mutate the DOM, while all that should've happened is the removal or addition of a single Component - something that should be instantaneous.

The workaround is to never change the underlying array/map references, and in turn not to use value-converters for repeaters (instead - always mutate the original instance that the View binds to). But then ValueConverters, one of the highlight features of Aurelia, loose a lot of their appeal. A task like filtering an array is pretty common in any app, even the simplest one such as the TodoMVC. It would be great if Aurelia tackled this problem systemically, because manually mutating an array for use-cases such as filtration requires a lot of manual hackery (keeping another "reference array" and manually multi-splicing the "filtered array" for each filtration/de-filtration action - huge pain in the butt!).

Sometimes it is also easier for the developer to simply provide a fresh array/map instead of mutating the old one; this is especially true when using immutable programming styles or observables. It would be great if Aurelia could simply figure out which elements got removed and which got added, without the need for the developer to reimplement this for every array.

Personally, this is a deal-breaker for me, since competitor frameworks do not have this issue - React, Cycle.js and any VirtualDom implementation such as snabbdom will always diff the values in order to optimally mutate the DOM - what matters are the individual components, not the iterable structure that holds them together.

EisenbergEffect:
@martingust @jdanyow Can we look into doing some more improvements in this area?
@niieani On this note, if you have a long list, you should really be using the ui-virtualization library so that you don't even make the browser render things that aren't' visible. Just install the plugin and switch your `repeat` to a `virtual-repeat`.

niieani:
[...]

PR #215, opened 2016-04-20
perf(repeat): intelligently reuse items from previous instance

niieani:
Fixes https://github.com/aurelia/templating-resources/issues/213.
Blocked on: https://github.com/aurelia/templating/pull/324.

Demo: https://gist.run/?id=af08bd464f72a809d07639b2d9f88a38

### DIFF
--- a/src/array-repeat-strategy.js
+++ b/src/array-repeat-strategy.js
@@ -1,4 +1,4 @@
-import {createFullOverrideContext, updateOverrideContexts} from './repeat-utilities';
+import {createFullOverrideContext, updateOverrideContexts, updateOverrideContext, indexOf} from './repeat-utilities';
 import {mergeSplice} from 'aurelia-binding';
 
 /**
@@ -20,16 +20,100 @@ export class ArrayRepeatStrategy {
   * @param items The new array instance.
   */
   instanceChanged(repeat, items) {
-    if (repeat.viewsRequireLifecycle) {
-      let removePromise = repeat.removeAllViews(true);
-      if (removePromise instanceof Promise) {
-        removePromise.then(() => this._standardProcessInstanceChanged(repeat, items));
-        return;
-      }
+    const itemsLength = items.length;
+    
+    // if the new instance does not contain any items, 
+    // just remove all views and don't do any further processing
+    if (!items || itemsLength === 0) {
+      repeat.removeAllViews(true);
+      return;
+    }
+    
+    const children = repeat.views();
+    const viewsLength = children.length;
+    
+    // likewise, if we previously didn't have any views, 
+    // simply make them and return
+    if (viewsLength === 0) {
       this._standardProcessInstanceChanged(repeat, items);
       return;
     }
-    this._inPlaceProcessItems(repeat, items);
+    
+    if (repeat.viewsRequireLifecycle) {
+      const childrenSnapshot = children.slice(0);
+      const itemNameInBindingContext = repeat.local;
+      const matcher = repeat.matcher();
+      
+      // the cache of the current state (it will be transformed along with the views to keep track of indicies)
+      let itemsPreviouslyInViews = [];
+      const viewsToRemove = [];
+      
+      for (let index = 0; index < viewsLength; index++) {
+        const view = childrenSnapshot[index];
+        const oldItem = view.bindingContext[itemNameInBindingContext];
+        
+        if (indexOf(items, oldItem, matcher) === -1) {
+          // remove the item if no longer in the new instance of items
+          viewsToRemove.push(view);
+        } else {
+          // or add the item to the cache list
+          itemsPreviouslyInViews.push(oldItem);
+        }
+      }
+      
+      let updateViews;
+      let removePromise;
+      
+      if (itemsPreviouslyInViews.length > 0) {
+        removePromise = repeat.removeViews(viewsToRemove, true);
+        updateViews = () => {
+          // update views (create new and move existing)
+          for (let index = 0; index < itemsLength; index++) {
+            const item = items[index];
+            const indexOfView = indexOf(itemsPreviouslyInViews, item, matcher, index);
+            let view;
+            
+            if (indexOfView === -1) { // create views for new items
+              const overrideContext = createFullOverrideContext(repeat, items[index], index, itemsLength);
+              repeat.insertView(index, overrideContext.bindingContext, overrideContext);
+              // reflect the change in our cache list so indicies are valid
+              itemsPreviouslyInViews.splice(index, 0, undefined);
+            }
+            else if (indexOfView === index) { // leave unchanged items
+              view = children[indexOfView];
+              itemsPreviouslyInViews[indexOfView] = undefined;
+            }
+            else { // move the element to the right place
+              view = children[indexOfView];
+              repeat.moveView(indexOfView, index);
+              itemsPreviouslyInViews.splice(indexOfView, 1);
+              itemsPreviouslyInViews.splice(index, 0, undefined);
+            }
+            
+            if (view) {
+              updateOverrideContext(view.overrideContext, index, itemsLength);
+            }
+          }
+          
+          // remove extraneous elements in case of duplicates, 
+          // also update binding contexts if objects changed using the matcher function
+          this._inPlaceProcessItems(repeat, items);
+        }
+      } else {
+        // if all of the items are different, remove all and add all from scratch
+        removePromise = repeat.removeAllViews(true);
+        updateViews = () => this._standardProcessInstanceChanged(repeat, items);
+      }
+      
+      if (removePromise instanceof Promise) {
+        removePromise.then(updateViews);
+      } else {
+        updateViews();
+      }
+    } else {
+      // no lifecycle needed, use the fast in-place processing
+      this._inPlaceProcessItems(repeat, items);
+    }
   }
 
   _standardProcessInstanceChanged(repeat, items) {
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -92,6 +92,7 @@ export class Repeat extends AbstractRepeater {
   */
   bind(bindingContext, overrideContext) {
     this.scope = { bindingContext, overrideContext };
+    this.matcherBinding = this._captureAndRemoveMatcherBinding();
     this.itemsChanged();
   }
 
@@ -101,6 +102,7 @@ export class Repeat extends AbstractRepeater {
   unbind() {
     this.scope = null;
     this.items = null;
+    this.matcherBinding = null;
     this.viewSlot.removeAll(true);
     this._unsubscribeCollection();
   }
@@ -204,11 +206,31 @@ export class Repeat extends AbstractRepeater {
       this.collectionObserver.subscribe(this.callContext, this);
     }
   }
+  
+  _captureAndRemoveMatcherBinding() {
+    if (this.viewFactory.viewFactory) {
+      const instructions = this.viewFactory.viewFactory.instructions;
+      const instructionIds = Object.keys(instructions);
+      for (let i = 0; i < instructionIds.length; i++) {
+        const expressions = instructions[instructionIds[i]].expressions;
+        if (expressions) {
+          for (let ii = 0; i < expressions.length; i++) {
+            if (expressions[ii].targetProperty === 'matcher') {
+              const matcherBinding = expressions[ii];
+              expressions.splice(ii, 1);
+              return matcherBinding;
+            }
+          }
+        }
+      }
+    }
+  }
 
   // @override AbstractRepeater
   viewCount() { return this.viewSlot.children.length; }
   views() { return this.viewSlot.children; }
   view(index) { return this.viewSlot.children[index]; }
+  matcher() { return this.matcherBinding ? this.matcherBinding.sourceExpression.evaluate(this.scope, this.matcherBinding.lookupFunctions) : null; }
 
   addView(bindingContext, overrideContext) {
     let view = this.viewFactory.create();
@@ -221,9 +243,17 @@ export class Repeat extends AbstractRepeater {
     view.bind(bindingContext, overrideContext);
     this.viewSlot.insert(index, view);
   }
+  
+  moveView(sourceIndex, targetIndex) {
+    this.viewSlot.move(sourceIndex, targetIndex);
+  }
 
   removeAllViews(returnToCache, skipAnimation) {
     return this.viewSlot.removeAll(returnToCache, skipAnimation);
+  }
+
+  removeViews(viewsToRemove, returnToCache, skipAnimation) {
+    return this.viewSlot.removeMany(viewsToRemove, returnToCache, skipAnimation);
   }
 
   removeView(index, returnToCache, skipAnimation) {
--- a/src/abstract-repeater.js
+++ b/src/abstract-repeater.js
@@ -39,6 +39,15 @@ export class AbstractRepeater {
   }
 
   /**
+   * Returns the matcher function to be used by the repeater, or null if strict matching is to be performed.
+   *
+   * @return {Function|null} The requested matcher function.
+   */
+  matcher() {
+    throw new Error('subclass must implement `matcher`');
+  }
+
+  /**
    * Adds a view to the repeater, binding the view to the
    * provided contexts.
    *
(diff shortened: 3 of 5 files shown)


EisenbergEffect:
@martingust @jdanyow Can you guys review?

src/array-repeat-strategy.js
```diff
@@ -20,16 +20,82 @@ export class ArrayRepeatStrategy {
   * @param items The new array instance.
   */
   instanceChanged(repeat, items) {
-    if (repeat.viewsRequireLifecycle) {
-      let removePromise = repeat.removeAllViews(true);
-      if (removePromise instanceof Promise) {
-        removePromise.then(() => this._standardProcessInstanceChanged(repeat, items));
+    const itemNameInBindingContext = repeat.local;
+    
+    const children = repeat.viewSlot.children;
```
martingust: Use the AbstractRepeater's [viewCount](https://github.com/aurelia/templating-resources/blob/master/src/repeat.js#L209) method here

src/array-repeat-strategy.js
```diff
@@ -20,16 +20,82 @@ export class ArrayRepeatStrategy {
...
+        }
+        
+        // remove also from the cache list:
+        itemsPreviouslyInViews.splice(indiciesToRemove[i], 1);
+      }
+      
+      // update views
+      Promise.all(rmPromises).then(() => {
+        let index = 0;
+        for (let item of items) {
```
martingust: We want to avoid `for..of` loops throughout the Aurelia codebase since the compiled code uses Symbols. Please change to a good old for-loop.

src/array-repeat-strategy.js
```diff
@@ -20,16 +20,82 @@ export class ArrayRepeatStrategy {
...
+          }
+          else { // move the element to the right place
+            view = children[indexOfView];
+            repeat.moveView(indexOfView, index);
+            itemsPreviouslyInViews.splice(indexOfView, 1);
+            itemsPreviouslyInViews.splice(index, 0, undefined);
+          }
+          
+          if (view) {
+            updateOverrideContext(view.overrideContext, index, itemsLength);
```
martingust: I think you will need to import this method from [repeat-utilities](https://github.com/aurelia/templating-resources/blob/master/src/repeat-utilities.js#L54)

src/repeat.js
```diff
@@ -221,6 +221,10 @@ export class Repeat extends AbstractRepeater {
     view.bind(bindingContext, overrideContext);
     this.viewSlot.insert(index, view);
   }
+  
+  moveView(sourceIndex, targetIndex) {
```
martingust: I would be great to have this method added to the [AbstractRepeater](https://github.com/aurelia/templating-resources/blob/master/src/abstract-repeater.js)

martingust:
@niieani Great work on this!! I added a few small notes that would be great if you could take a look at. If you can add tests for this as well it would be great. You can take a look at the integrations tests [here](https://github.com/aurelia/templating-resources/blob/master/test/repeat-integration.spec.js). Let me know if any questions.

Thanks!

src/array-repeat-strategy.js
niieani: Indeed! I've had it, but forgot to add that in the cleaned up version (it's on the gist.run demo though).

src/repeat.js
niieani: OK, will add there too.

src/array-repeat-strategy.js
niieani: OK.

src/array-repeat-strategy.js
niieani: Got it.

niieani: I think you mean `views()` method, not `viewCount()` :-)

niieani:
@martingust I'm having problems running the tests on Aurelia's `master` of templating-resources.
Tests seem to fail even without any of my changes.
Here's what happens:

```shell
➜  templating-resources git:(perfwork) ✗ gulp test
[10:50:52] Warning: gulp version mismatch:
[10:50:52] Global gulp is 3.9.1
[10:50:52] Local gulp is 3.9.0
[10:50:53] Using gulpfile ~/Downloads/GitHub/aurelia/templating-resources/gulpfile.js
[10:50:53] Starting 'test'...
21 04 2016 10:50:56.396:INFO [karma]: Karma v0.13.19 server started at http://localhost:9876/
21 04 2016 10:50:56.404:INFO [launcher]: Starting browser Chrome
21 04 2016 10:50:57.622:INFO [Chrome 51.0.2704 (Mac OS X 10.11.4)]: Connected on socket /#khWOppiHhBrMJKfnAAAA with id 56942052
Chrome 51.0.2704 (Mac OS X 10.11.4) ERROR
  Uncaught TypeError: Cannot read property 'firstChild' of undefined
  at /Users/nixin/Downloads/GitHub/aurelia/templating-resources/jspm_packages/npm/aurelia-task-queue@1.0.0-beta.1.1.0/aurelia-task-queue.js:40
Chrome 51.0.2704 (Mac OS X 10.11.4): Executed 37 of 108 ERROR (6.691 secs / 6.66 secs)
[10:51:05] 'test' errored after 12 s
[10:51:05] Error: 1
    at formatError (/usr/local/lib/node_modules/gulp/bin/gulp.js:169:10)
    at Gulp.<anonymous> (/usr/local/lib/node_modules/gulp/bin/gulp.js:195:15)
    at emitOne (events.js:90:13)
    at Gulp.emit (events.js:182:7)
    at Gulp.Orchestrator._emitTaskDone (/Users/nixin/Downloads/GitHub/aurelia/templating-resources/node_modules/orchestrator/index.js:264:8)
    at /Users/nixin/Downloads/GitHub/aurelia/templating-resources/node_modules/orchestrator/index.js:275:23
    at finish (/Users/nixin/Downloads/GitHub/aurelia/templating-resources/node_modules/orchestrator/lib/runTask.js:21:8)
    at cb (/Users/nixin/Downloads/GitHub/aurelia/templating-resources/node_modules/orchestrator/lib/runTask.js:29:3)
    at removeAllListeners (/Users/nixin/Downloads/GitHub/aurelia/templating-resources/node_modules/karma/lib/server.js:333:7)
    at Server.<anonymous> (/Users/nixin/Downloads/GitHub/aurelia/templating-resources/node_modules/karma/lib/server.js:344:9)
    at Server.g (events.js:273:16)
    at emitNone (events.js:85:20)
    at Server.emit (events.js:179:7)
    at emitCloseNT (net.js:1529:8)
    at _combinedTickCallback (internal/process/next_tick.js:71:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

src/array-repeat-strategy.js
```diff
@@ -20,16 +20,82 @@ export class ArrayRepeatStrategy {
...
+    const children = repeat.viewSlot.children;
+    let itemsPreviouslyInViews = children.map(view => view.bindingContext[itemNameInBindingContext]);
+    
+    let itemsLength = items.length;
+    
+    // remove items no longer in the instance
+    if (itemsPreviouslyInViews.length > 0) {
+      let indiciesToRemove = [];
+      
+      itemsPreviouslyInViews.forEach((oldItem, index) => {
```
jdanyow: Please replace the `.forEach` with a regular for-loop.  Will reduce the memory pressure a bunch by eliminating all the closures that are created.

src/array-repeat-strategy.js
```diff
@@ -20,16 +20,82 @@ export class ArrayRepeatStrategy {
...
   */
   instanceChanged(repeat, items) {
-    if (repeat.viewsRequireLifecycle) {
-      let removePromise = repeat.removeAllViews(true);
-      if (removePromise instanceof Promise) {
-        removePromise.then(() => this._standardProcessInstanceChanged(repeat, items));
+    const itemNameInBindingContext = repeat.local;
+    
+    const children = repeat.viewSlot.children;
+    let itemsPreviouslyInViews = children.map(view => view.bindingContext[itemNameInBindingContext]);
```
jdanyow: Please remove the `.map`.  Will reduce the memory pressure a bunch by eliminating all the closures that are created.

niieani: @jdanyow Should I replace it with a regular for-loop?

niieani:
I've corrected all the problems pointed out by @jdanyow and @martingust, added 13 tests to this one and rebased. Thanks to the removal of the `map`ping suggested by @jdanyow I've additionally managed to remove one more for loop by [merging the item listing with removal actions](https://github.com/niieani/templating-resources/blob/master/src/array-repeat-strategy.js#L43-L55). 
The related pull request at https://github.com/aurelia/templating/pull/324 has now 3 tests too. I think they should cover all new use-cases. If you think everything is now ok, this should be ready to merge.
CC: @EisenbergEffect

martingust:
Awesome stuff, big thanks @niieani!

niieani:
Few more things to think about once this is merged:

Note that because of how [`callSplices()` method](https://github.com/aurelia/binding/blob/master/src/array-change-records.js#L148) works the new optimizations won't apply to modifications on the same instance of the array, e.g. 
```js
this.array = [9, 8, 7, 6, 5, 4, 3, 2, 1];
this.array.reverse();
```
Since this operates on the same Array, Aurelia uses the [splices mechanism](https://github.com/aurelia/binding/blob/master/src/array-change-records.js#L148) to modify the Views.

So in fact, given my code is merged, a developer would get better performance with the following code:
```js
this.array = [9, 8, 7, 6, 5, 4, 3, 2, 1];
this.array = this.array.slice(0).reverse();
```

I don't know what's the best way to make Aurelia use my View diffing mechanism in such cases, because once we're already inside the RepeatStrategy's `instanceMutated()`, there's no way to differentiate the "reset splices" from, say, `array.slice(0, 1)`. This would thus involve modifying `aurelia-binding` so that these special `reset` operations add a dedicated field that the RepeatStrategy could understand.

Here's a demo of this problem:
https://gist.run/?id=2c6423601fbecde7ef62b4087564576d

Clicking "swapArrays" won't cause any of the repeat Views to be unbound & bound and recreated, but clicking `.reverse()` does.

One more outstanding issue after this is merged is to do with immutability, brought up by @jdanyow. It might make sense to extend this "intelligent View reuse" behavior to completely new objects, provided the developer provides a way to compare them against each other (e.g. an ID property). In such a case we'd simply update their `bindingContext`, in addition to the reuse-rearrange logic. Building on the work I've already done on this, I believe it would be a very simple addition to the repeater code. 

Then there is also the case of adapting the code from the ArrayRepeatStrategy (which I've worked on) to other RepeatStrategies, including `Map` and `Set`.

EisenbergEffect:
@niieani Fantastic work. @martingust @jdanyow Are you both happy with this? If so, let me know and I'll merge both PRs. @niieani Really excellent contribution. I'd love to see you continue to brainstorm on this with @martingust and @jdanyow It sounds like we can continue to make the repeater smarter in the future. This is definitely a tremendous improvement for the community. Keep it up!

martingust:
Very happy. This is definitely something we should continue to improve upon for the other strategies, mutation, also same logic should be implemented in UI-Virtualization.

src/array-repeat-strategy.js
```diff
@@ -20,16 +20,79 @@ export class ArrayRepeatStrategy {
...
   instanceChanged(repeat, items) {
-    if (repeat.viewsRequireLifecycle) {
-      let removePromise = repeat.removeAllViews(true);
-      if (removePromise instanceof Promise) {
-        removePromise.then(() => this._standardProcessInstanceChanged(repeat, items));
+    // if the new instance does not contain any items, 
+    // just remove all views and don't do any further processing
+    if (!items || items.length === 0) {
+      repeat.removeAllViews(true);
+      return;
```
niieani: I just noticed this chunk wasn't necessary anymore (no need to process, because there are no views to add):
```js
      if (removePromise instanceof Promise) {
        removePromise.then(() => this._standardProcessInstanceChanged(repeat, items));
      }
```
So I removed that too, making the code even cleaner.

jdanyow:
@niieani I just pulled this down and gave it a test drive.  It's looking really good. I think we can merge this as soon as the tests are passing and it's working with dbmonster.

1. Need to implement abstractrepeater's `moveView` in the repeat (I bet you have this locally already)
2. To make the test suite pass you'll need to pull down the latest version of binding, build it, and copy the contents of it's dist/commonjs folder over the contents of the templating-resources/jspm_packages/npm/aurelia-binding@xxxx folder.  The repeat has some tests related to shifting/popping an empty array that will fail unless the latest binding build is used.
3. dbmonster: lot of errors occur when using these changes with dbmonster.  dbmonster is an example of rapidly replacing an array instance with a new instance (with new contents).  Need to figure out why this breaks the repeat.  I have a feeling it's something to do with the remove promises.  The new logic effectively kills the "inPlaceProcessItems" optimizations by removing views.  There's an updated dbmonster to test with [here](https://github.com/jdanyow/aurelia-dbmonster/).

Optional- this could go in a separate PR:

We had talked about handling data from the server that might be different object instances but effectively the same data [here](https://github.com/aurelia/templating-resources/issues/213#issuecomment-211930267).  We have a pattern that's used in areas of the binding system that might compliment your repeat improvements by solving this part of the issue.  In select, checkbox and radio binding (pickers) developers can supply an optional [`matcher`](https://github.com/aurelia/binding/blob/master/src/select-value-observer.js#L65) function.  Would it make sense to do the same thing here?  Then you can use the matcher function instead of a strict equality check.

thanks again for all the work on this.  We're looking forward to getting this in.

niieani:
@jdanyow 

1. duh! A forgot to include this change when rebasing, I had it in the previous version.
2. Will test. The test suite was passing for me except for those two tests which didn't even work with the `master` and without any of my changes. But I'll try with the updated aurelia-binding as you say.
3. I'll take a look at dbmaster now.

The `matcher` function sounds very reasonable indeed - it can fall back to strict checking if no matcher is present. I'll see how difficult it would be to include in this PR, but as you say it might make more sense to first get this one working with dbmonster and then work on expanding it.

niieani:
1. Fixed dbMonster (was a race condition, needs another new method for the `ViewSlot`, pull request here: https://github.com/aurelia/templating/pull/325)
2. Further optimization of adding/moving/removing part.
3. Added the `matcher` functionality, along with tests.

Demo of the matcher: https://gist.run/?id=e572641799a00f5eccda510d61f9ac89

EisenbergEffect:
Really fantastic work @niieani 
@jdanyow Please review :)

jdanyow:
really nice PR- thanks @niieani!

EisenbergEffect:
SWEET!
@jdanyow How is db monster doing after this? Was it affected at all? Are we still showing the same perf as before?

jdanyow:
Yep same as before- @niieani's pr improves a different use-case- arguably a more realistic one

EisenbergEffect:
Great. Mostly with dbmonster, it's about marketing....so just want to make sure we didn't have a regression there. Very excellent stuff here :)

niieani:
Glad to help!
Indeed, I find DbMonster's usage really contrived too (when would anybody realistically, in any kind of an application replace all the items multiple times per second?). 

Also, it'll be useful to mention the new `matcher` functionality in the docs.

@EisenbergEffect Since you're interested in numbers for marketing, it was possible to run DbMonster with the viewCycle-optimization off, forcing it to run against my new code (or the old unoptimized code). Before my optimization, I was getting around 1-1.5 fps, now I'm getting about 10 fps on my Macbook Pro, so that's a factor of 8-10 improvement simply thanks to the reuse of unchanged items (using DbMonster's default 50%). Mind you, I wasn't doing a proper benchmark, only playing around when debugging, so do take the numbers with a grain of salt.

brandonseydel:
@niieani Have you used this in conjunction with the virtual repeater to do any benchmarks on the DBMonster implementation.  I have been following and pulling your implementations and testing myself throughout the progression of this enhancement, but I have yet to test with the virtual repeater.

niieani:
@brandonseydel virtual repeater will need to be updated separately to match the new behavior. I've given quite a bit of time to this now, and I'm currently working on my aurelia-cycle plugin. I didn't analyze how exactly virtualized-ui works, so I'm not sure what the performance benefits will be there. I guess it'll depend on the underlying data and on the number of visible elements.

PR #234, opened 2016-06-18
test(repeat): use the component tester

martingust:
@EisenbergEffect This requires `testing` to be released with [this](https://github.com/aurelia/testing/pull/14).

@jdanyow There are two tests failing I couldn't get to pass, [this](https://github.com/aurelia/templating-resources/blob/master/test/hide.spec.js#L36) for `hide` and [this](https://github.com/aurelia/templating-resources/blob/master/test/show.spec.js#L36) for `show`. Looks like same issue, both giving me `Expecting undefined to be true`. Any ideas?

### DIFF
--- a/config.js
+++ b/config.js
@@ -17,7 +17,11 @@ System.config({
   map: {
     "aurelia-binding": "npm:aurelia-binding@1.0.0-beta.2.0.5",
     "aurelia-dependency-injection": "npm:aurelia-dependency-injection@1.0.0-beta.2.1.0",
+    "aurelia-event-aggregator": "npm:aurelia-event-aggregator@1.0.0-beta.2.0.0",
+    "aurelia-framework": "npm:aurelia-framework@1.0.0-beta.2.0.0",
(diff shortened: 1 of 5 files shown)


jdanyow:
There's a few calls to `withResources()` with no args.  I think this method on the component tester only accepts an array.

After replacing those with `withResources([])` there's an error [here](https://github.com/aurelia/testing/blob/master/src/component-tester.js#L58) because host.firstElementChild is null.

martingust:
@jdanyow I've created a PR in testing that removes the need for providing resources, that will need to be released in order for these tests to pass.

EisenbergEffect:
@martingust I can merge that if you can revert the change to the eslint.json file

jdanyow:
@martingust I think all the tests are broken- repeat included.  Hit karma's debug button and check the console

EisenbergEffect:
I only had two failing tests. Make sure all your deps are up to date.

martingust:
Same here, also two failing in latest CI build https://circleci.com/gh/aurelia/templating-resources/14

PR #261, opened 2016-11-07
fix scss import as require in html templates

florinspataru1:
This is because the sccs import is failing. I am working within a team and we are using webpack with aureliajs. We encountered a big issue while using scss files and this fixes the issue.

### DIFF
--- a/dist/amd/array-repeat-strategy.js
+++ b/dist/amd/array-repeat-strategy.js
@@ -150,10 +150,10 @@ define(['exports', './repeat-utilities', 'aurelia-binding'], function (exports, 
 
       if (repeat.__queuedSplices) {
         for (var i = 0, ii = splices.length; i < ii; ++i) {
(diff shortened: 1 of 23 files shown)


jdanyow:
@florinspataru1 I think this is something that might belong in your main.js.  Please amend the PR to only include `src` files while you wait for @EisenbergEffect / @niieani to review.

Thanks!

niieani:
I think @EisenbergEffect wanted to expose a method for users that would allow aliasing other extensions with `.css` or `.html` with ease. I think that's a better, generalized solution, than hardcoding `.sass`.

EisenbergEffect:
I'm going to close this for a few of reasons:

* I don't think this will work due to it not being able to either change the file extension or convert the scss (one of the other).
* There were a bunch of unrelated changes made to the dist folder as part of this PR.
* We would rather have an api to enable supporting arbitrary extensions if we can.

PR #263, opened 2016-11-15
feat(BindingBehavior): Support self binding behavior for event

bigopon:
aurelia/templating-binding#101

Tests are still failing (throttle, debounce, self) as aurelia-binding dependency is not updated

Usage (& test):
```html
<div click.delegate="doSomething($event.target.tagName.toLowerCase() === 'div') & self">
  <button click.delegate="wontdo">Clicking me won't trigger my div parent listener</button>
</div>
```

Where this can be useful:

Suggest we have a modal component with following markup:

```html
<div class="overlay">
  <div class="content-wrapper">
    <!-- ...... -->
  </div>
</div>
```

Without self binding behavior we need to add some extra stuff or separate dom path for `div.overlay` to perform action close on click the overlay.

With self binding behavior:

```html
<div class="overlay" click.delegate='close() & self'>
  <div class="content-wrapper">
    <!-- ...... -->
  </div>
</div>
```


EisenbergEffect:
Can you explain this feature further? @jdanyow Can you review?

bigopon:
In this example

```html
<panel>
  <header mousedown.delegate='onMouseDown()'>
    <button class='settings'></button>
    <button class='close'></button>
  </header>
</panel>
```

Use case is I want when I mouse down on header, it start listening to `mousemove` to do drag. In the above markup, it requires to check if `event.target === header` before you do something else to make sure you don't start listening to `mousemove` when mouse down on the buttons.

* Rewritten with self binding behavior:

```html
<panel>
  <header mousedown.delegate='onMouseDown() & self'>
    <!-- mouse down on any of these buttons won't call onMouseDown -->
    <button class='settings'></button>
    <button class='close'></button>
  </header>
</panel>
```

The `self` binding behavior will help remove event stuff of of your business/ component logic so `onMouseDown` is ensured to only fire when user actually mouse down on header itself, not its descendants.

src/self-binding-behavior.js
```diff
@@ -0,0 +1,22 @@
...
+function handleSelfEvent(event) {
+  let target = findOriginalEventTarget(event);
+  if (this.target !== target) return;
+  this.originalCallSource(event);
+}
+
+export class SelfBindingBehavior {
+  bind(binding, source) {
+    if (!binding.callSource || !binding.targetEvent) throw new Error('Self binding behavior only supports event.');
+    binding.originalCallSource = binding.callSource;
```
jdanyow: instead of `originalCallSource` (which might conflict with someone else's binding behavior logic) try using `selfCallSource` or something else that is specific to this binding behavior.

src/self-binding-behavior.js
```diff
@@ -0,0 +1,22 @@
...
+    binding.callSource = handleSelfEvent;
+  }
+
+  unbind(binding, source) {
+    this.callSource = this.originalCallSource;
```
jdanyow: shouldn't `this.` be `binding.` here?

src/self-binding-behavior.js
```diff
@@ -0,0 +1,22 @@
...
+    this.originalCallSource = null;
+  }
```
jdanyow: `this.` -> `binding.`

test/debounce-binding-behavior.spec.js
```diff
@@ -168,7 +169,7 @@ describe('DebounceBindingBehavior', () => {
       bindingEngine.observerLocator.eventManager,
       'mousemove',
       bindingEngine.parseExpression(`handleMouseMove($event) & debounce:${delay}`),
-      false,
+      delegationStrategy.none,
       false,
```
jdanyow: does this belong in another PR?

test/throttle-binding-behavior.spec.js
```diff
@@ -143,7 +144,7 @@ describe('ThrottleBindingBehavior', () => {
       bindingEngine.observerLocator.eventManager,
       'mousemove',
       bindingEngine.parseExpression(`handleMouseMove($event) & throttle:${delay}`),
-      false,
+      delegationStrategy.none,
       false,
```
jdanyow: does this belong in another PR

bigopon: Hi, it was already merged but I don't know how to update my fork. Can you help please? I tried but couldn't find any clue why.

EisenbergEffect:
@bigopon Thanks for the explanation. That sounds great.

test/throttle-binding-behavior.spec.js
bigopon: Great, I read about rebase but not sure if it was the right thing to do, git documentation was quite a puzzle for me plus I'm still newbie. Tks.

Updates:
- Rebased
- Added test for all `delegationStrategy`
- Fixed unbind

bigopon:
@jdanyow I've updated all deps. Tests are passing now

jdanyow:
@bigopon this is looking really good. One last thing- could you add a section to this doc that explains how to use this feature?

https://github.com/aurelia/binding/blob/master/doc/article/en-US/binding-binding-behaviors.md

bigopon:
@jdanyow sure, doing. But it seems to be more challenging than the writing code 😢

PR #279, opened 2017-03-17
feat(templating): support popular transpilers

### LINKED ISSUE #277: Require stylesheets other than CSS
Support for CSS in `<require from="style.css">` is currently hardcoded here:
https://github.com/aurelia/templating-resources/blob/fe7a836504a0ee0fdc8e53480cd9f3d6c9dc4f7e/src/aurelia-templating-resources.js#L66

Today lots of people use pre-processors, like LESS or SASS.

In a webpack build, `<require from="style.less">` would do just what you expect (great!)... except that it would fail at runtime without some hardcore configuration because the module extension is not `.css`.

A quick easy fix that would help 90% of our users can be to hardcode a few more values, I suggest adding `.scss`, `.sass`, `.less`. That's very reasonable defaults.

Of course going the extra mile would be exposing a configuration api for exotic preprocessors, but that's a lot more work for less benefits (few users, if any, that could use other work-arounds).

The fix should also be applied in HMR, where this is duplicated CC @niieani

jods4:
LESS and SASS extensions can now be loaded as `<style>` by `<require from=style.less>` in views.
Of course, content should previously have been properly compiled to CSS.

Fixed #277

### DIFF
--- a/src/aurelia-templating-resources.js
+++ b/src/aurelia-templating-resources.js
@@ -68,7 +68,7 @@ function configure(config) {
       return { [address]: _createCSSResource(address) };
     }
   };
-  ['.css', '.less', '.sass', '.scss'].forEach(ext => viewEngine.addResourcePlugin(ext, styleResourcePlugin));
+  ['.css', '.less', '.sass', '.scss', '.styl'].forEach(ext => viewEngine.addResourcePlugin(ext, styleResourcePlugin));
 }
 
 export {
--- a/src/aurelia-templating-resources.js
+++ b/src/aurelia-templating-resources.js
@@ -63,11 +63,12 @@ function configure(config) {
   configureHtmlResourcePlugin(config);
 
   let viewEngine = config.container.get(ViewEngine);
-  viewEngine.addResourcePlugin('.css', {
-    'fetch': function(address) {
+  let styleResourcePlugin = {
+    fetch(address) {
       return { [address]: _createCSSResource(address) };
     }
-  });
+  };
+  ['.css', '.less', '.sass', '.scss'].forEach(ext => viewEngine.addResourcePlugin(ext, styleResourcePlugin));
 }
 
 export {


jods4:
@niieani can you do the same for HMR? I know the same logic is duplicated somewhere.

niieani:
I can do the same for HMR, but could you `export const styleResourceExtensions = ['.css', '.less', '.sass', '.scss']` here (or something along the lines)? This way, whenever we add or change the list, we'd only have to do it in one place. [This is the place](https://github.com/aurelia/hot-module-reload/blob/master/src/aurelia-hot-module-reload.ts#L46-L55) where it's got to be done for HMR.

jods4:
@niieani I don't think it's a good idea.

Don't get me wrong: I agree the code is not pretty and hate that there is so much duplication between templ-res and HMR and that any change here must be reflected over there. It's clearly a bad smell.

But there are a handful of reasons why I don't fancy exporting this array, the main ones being:
- `hot-module-reload` should not take a dependency on `templating-resources`. The latter is an optional Aurelia component, that is included when you do `use.standardConfiguration()` but is not mandatory. HMR should work without it, otherwise it creates a significant discrepency between a debug build and the production one.
- Ideally I think we would want a clean configuration API with reasonable defaults, and introspection that can be use by HMR and others. Probably in `aurelia-framework`, something like: `aurelia.use.stylesheets('.less')`. In the meantime, exporting this array means some users might take a dependency on it, maybe cleverly modifying it before Aurelia starts. Once exposed it's a breaking change to remove :(
- There is so much duplication here that exposing this array is a joke. `hmr-css-resource.ts` is a 100 lines copy (with 2 minor changes) of `css-resource.js`. Exporting this array is really the cherry at the top of the cake.

As I wrote in the issue, I see this as a dirty internal quick-fix to enable requiring styles for more users. I wouldn't change the layering of libs or the exposed surface before we have a better designed, comprehensive solution.

😕

niieani:
OK, makes sense.

bigopon:
Thanks @jods4 @niieani @EisenbergEffect . Now my html only components are truely html only

PR #280, opened 2017-03-19
feat(templating): else custom attribute

### LINKED ISSUE #171: If-Else pattern, "let" element
I'm finding myself constantly writing code that seems like an anti-pattern to me:

``` html
<div if.bind="property | transform | something:param1 | pluck:'blah' & signal:'some-signal'">
   (something)
</div>
<div if.bind="!(property | transform | something:param1 | pluck:'blah') & signal:'some-signal'">
   (something else)
</div>
```

This is clearly an If-Else pattern, but here it feels like an anti-pattern, because of:
1. repetition of code (not DRY)
2. double execution of the whole transformation chain

Is there a better way to do this that I'm not aware of? How could an `else` be implemented as a custom attribute? Perhaps something like:

``` html
<template if.bind="property | transform | something:param1 | pluck:'blah' & signal:'some-signal'">
  <true>
    (true content)
  </true>
  <false>
    (false content)
  </false>
</template>
```

Thoughts?

jods4:
This PR introduces an `else` custom attribute that can be used directly following a `if` binding:
```html
<div if.bind="whatever">Whatever is truthy!</div>
<div else>No luck</div>
```

It also introduces an option to `if`.
- `condition` is the default option, as above.
- `swap: "after"|"with"|"before"` coordinates the animations (if any) between the `if` and the `else` templates.
```html
<div if="condition.bind: whatever; swap: with">Yes</div>
<div else class="au-animate">No</div>
```

The advantages of this are:
- Simpler, more intuitive to use. Especially when `condition` is a complex expression.
- The pair only subscribes for changes once.
- Coordinated animations.

⚠️ **Don't merge this before aurelia/templating#537 and remember to bump the minimum dependency!!** Otherwise using `if` results in infinite recursion.

Tests cases will come later but the code is ready for review.

Fixes #171 

CC @EisenbergEffect @jdanyow for review
CC @MeirionHughes because you asked me to

### DIFF
--- a/src/if.js
+++ b/src/if.js
@@ -87,7 +87,7 @@ class IfCore {
 @inject(BoundViewFactory, ViewSlot)
 export class If extends IfCore {
   @bindable({ primaryProperty: true }) condition: any;
-  @bindable swap: "before"|"with"|"after";
+  @bindable swapOrder: "before"|"with"|"after";
 
   /**
   * Binds the if to the binding context and override context
@@ -132,7 +132,7 @@ export class If extends IfCore {
   }
 
   _swap(remove, add) {
-    switch (this.swap) {
+    switch (this.swapOrder) {
       case "before":
         return Promise.resolve(add._show()).then(() => remove._hide());
       case "with":
--- a/src/if.js
+++ b/src/if.js
@@ -1,82 +1,53 @@
-import {BoundViewFactory, ViewSlot, customAttribute, templateController} from 'aurelia-templating';
+import {BoundViewFactory, ViewSlot, bindable, customAttribute, templateController} from 'aurelia-templating';
 import {inject} from 'aurelia-dependency-injection';
+import {DOM} from 'aurelia-pal';
 
-/**
-* Binding to conditionally include or not include template logic depending on returned result
-* - value should be Boolean or will be treated as such (truthy / falsey)
-*/
-@customAttribute('if')
-@templateController
-@inject(BoundViewFactory, ViewSlot)
-export class If {
-  /**
-  * Creates an instance of If.
-  * @param {BoundViewFactory} viewFactory The factory generating the view
-  * @param {ViewSlot} viewSlot The slot the view is injected in to
-  */
+class IfCore {
   constructor(viewFactory, viewSlot) {
     this.viewFactory = viewFactory;
     this.viewSlot = viewSlot;
-    this.showing = false;
     this.view = null;
     this.bindingContext = null;
     this.overrideContext = null;
+    // If the child view is animated, `value` might not reflect the internal 
+    // state anymore, so we use `showing` for that. 
+    // Eventually, `showing` and `value` should be consistent.
+    this.showing = false;
   }
 
-  /**
-  * Binds the if to the binding context and override context
-  * @param bindingContext The binding context
-  * @param overrideContext An override context for binding.
-  */
   bind(bindingContext, overrideContext) {
     // Store parent bindingContext, so we can pass it down
     this.bindingContext = bindingContext;
     this.overrideContext = overrideContext;
-    this.valueChanged(this.value);
   }
 
-  /**
-  * Invoked everytime value property changes.
-  * @param newValue The new value
-  */
-  valueChanged(newValue) {
-    if (this.__queuedChanges) {
-      this.__queuedChanges.push(newValue);
+  unbind() {
+    if (this.view === null) {
       return;
     }
 
-    let maybePromise = this._runValueChanged(newValue);
-    if (maybePromise instanceof Promise) {
-      let queuedChanges = this.__queuedChanges = [];
+    this.view.unbind();
 
-      let runQueuedChanges = () => {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@jods4 By any chance did you investigate using a separate attribute for swap-mode? Since that's what we have for router-view and compose, I can see a certain nicety in enabling the same attribute name.

jods4:
@EisenbergEffect I did, that was my first idea.
`router-view` and `compose` are custom elements so the situation is a bit different.
`if` is a custom attribute and a `templateController`. So other attributes like `if-swap-mode` end up on the instantiated template _inside_ the `if` by default and are not part of the controller.

If there is a reasonable way to make `if-swap-mode` attribute work in conjonction with `if` then I'd gladly update the PR. Can you help with directions?

EisenbergEffect:
Ah, yes. You are right. That would be an issue. I'm ok with your approach then. Could we rename `swap` to `swap-order` for consistency in naming?

jods4:
@EisenbergEffect renamed to `swapOrder`.

Ixonal:
Don't mean to intrude, but was kind of curious about a couple aspects of this. Does this work properly as an else-if? For example, 

```html
<div if.bind="isFoo">Foo</div>
<div else if.bind="isBar">Bar</div>
<div else>Baz</div>
```

Also, would it be appropriate to allow some number of nodes between the else and the if, but always have an "else" refer to the first prior "if" in the DOM under the same parent? For example,

```html
<div if.bind="isFoo">Foo</div>
<div>Bar</div>
<div else>Baz</div>
```

MeirionHughes:
That is a good point to bring up. both `if` and `else` are template controllers, I think. I've been led to believe having multiple template controllers on the same element is a bad thing due to the order of execution being undefined.

jods4:
Having multiple template controllers on the same tag does not work. So combining `else` +`if` won't work.

I'm covering the main use case with this PR. There's always room for further improvement. If there's a real demand in the future maybe we could chain the `else` bindings:
```html
<div if.bind="a === 1"></div>
<div else.bind="b === 2"></div>
<div else></div>
```

As for your second example, I don't want to encourage it. I see `else` as a simple way to have alternative content in the same spot. For complex layouts in unrelated parts I would encourage a more structured VM.
That said... it _does_ work. `else` bindings work **as long as there is no Aurelia controller between itself and a `if` previous sibling**.

Ixonal:
I could see a demand for else-if functionality, but it's all up to you, man. What you've done already is much appreciated.

huochunpeng:
Nice feature, but a merge without test coverage seems rushing to me, especially when it breaks existing test.

It looks like there are few edge cases to be tested.

jods4:
Sorry, haven't got time to create the missing unit tests right now but I'll do so.
It doesn't mean it's untested as you can see in the GIF above.

[quote]especially when it breaks existing test.
It looks like there are few edge cases to be tested.
[/quote]

I was not aware of a failing test, which one?
As I'm going to write unit tests for this I'm interested in the edge cases that you think are worthy of inclusion?

huochunpeng:
If you see the commits https://github.com/aurelia/templating-resources/commits/master
You got a red cross on the merge.

Run the test locally, it breaks existing `if.spec.js`.

Speaking about the edge case, I have not tried this new feature, nor read the code. But from what I read here, I guess the `else` attr needs to look up previous dom element in order to find matching `if` attr.

There are two features in aurelia I guess you need to check.
1. `if` and `else` directly on `<template>` element. Remeber template element doesn't create a DOM element.
2. `if` and `else` directly on a containerless component. `@containerless() export class ContainerLessComponent {...}`

Here is a list in my mind. Only write with `<template`, need to repeat all test cases on `<container-less-component`

```
<template if="condition">...</template>
<template else>...</template>

<div if="condition">...</div>
<template else>...</template>

<template if="condition">...</template>
<div else>...</div>
```

Here is a tricky one for you, I have not tried it, don't know the outcome. The `else` should match reverse of `condition1`, not `condition2`.
```
<template if="condition1">
  <div if="condition2">...</div>
</template>
<div else>...</div>
```

jods4:
> If you see the commits https://github.com/aurelia/templating-resources/commits/master
You got a red cross on the merge.
Run the test locally, it breaks existing if.spec.js.

If you read the original post, you'll see that I wrote a warning.
When it was committed this code shed light on a templating bug that has been fixed since then.

huochunpeng:
Thanks for the explanation.

niemyjski:
+1 for else if :)

ejsmith:
I went to go try and see how you use this new binding after seeing the blog post and I noticed 2 things. Searching for "if" in the docs doesn't return any results and the docs don't show any examples of using "else".

jods4:
@niemyjski This PR being now merged, I think a feature request for `elseif` or similar syntax is better served by a new issue that people can upvote and that we can track!

@ejsmith I did not update the hub docs, maybe we need an issue for that as well 🤔

PR #290, opened 2017-07-13
proposed fix for issue 287

loaded02:
Fix propsed for wrong map indexes after map content changed. See #287 .
I also added some tests for map-repeat-strategy. 

The 'should correctly handle adding items after clear (issue 287)' test is somehow always ok. Even without the fix. I think this has something to do with wrong override context being bound to the viewSlot.children. But this is just an error in the test which I dont understand. When I step through map-repeat-strategy.js and updateOverrideContext() in repeat-utilities.js the index is now being set right! 

@jdanyow Hope that helps :)

### DIFF
--- a/src/map-repeat-strategy.js
+++ b/src/map-repeat-strategy.js
@@ -47,6 +47,7 @@ export class MapRepeatStrategy {
     let ii;
     let overrideContext;
     let removeIndex;
+    let addIndex;
     let record;
     let rmPromises = [];
     let viewOrPromise;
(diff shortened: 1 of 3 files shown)


EisenbergEffect:
@AStoker @mbroadst Can you review this? It looks similar to the other fix, but just want a second pair of eyes here. Thanks!

AStoker:
I'm not the most up to date on how the templating engine runs, but the change looks good to me, and the tests seem to make sense :+1:

mbroadst:
I'm basically in the same boat here! The code change is minimal and makes sense (without running it), and the test coverage seems more than adequate for the proposed fix. LGTM

Issue #273, opened 2017-02-06
focus.one-time unexpected behaviour


RomkeVdMeulen:
**I'm submitting a bug report**

* **Library Version:**
1.2.0

**Please tell us about your environment:**
* **Operating System:**
Linux (Ubuntu)

* **Node Version:**
6.9.4

* **NPM Version:**
3.10.10

* **JSPM**
JSPM 0.16.48

* **Browser:**
Chrome 55

* **Language:**
TypeScript 2.1.4

**Current behavior:**
If I want an element focused on render, but I'm not interested in keeping track of its focus state after that, I might do something like this:

```html
<button focus.one-time="true">My Button</button>
```

In this case, attaching event listeners to the element only results in needless overhead.

**Expected/desired behavior:**
The Focus custom attribute should check if the binding mode is two-way and if not, should not attach event listeners to the element.

RomkeVdMeulen:
While we're on the subject: I would expect the default binding mode of focus to be one-way. I usually bind it to an expression that cannot be assigned to:

```html
  <a href="#" focus.bind="item === defaultItem">${item.name}</a>
```

If focus changes and comes back to this link, I'll get an error:

```
Uncaught Error: Binding expression "item === defaultItem" cannot be assigned to.
```

I can't speak for other developers, but I expect the most common use-case for the focus attribute to set initial focus, not to keep a property updated on whether an item is focused or not.

Alexander-Taran:
attach focus from dialog is what you want..
question is how many developers thought about using focus.one-way

PR #291, opened 2017-07-20
refactor(Focus): Use handleEvent instead of arrow functions

bigopon:
This PR replaces 2 arrow functions: `focusListener` & `blurListener` by `handleEvent`, leveraging event target interface for lower memory consumption.

### DIFF
--- a/src/focus.js
+++ b/src/focus.js
@@ -69,19 +69,9 @@ export class Focus {
   }
 
   handleEvent(e) {
-    switch (e.type) {
-    case 'focus': this.handleFocus(e); break;
-    case 'blur': this.handleBlur(e); break;
-    default: break;
-    }
-  }
-
-  handleFocus(e) {
-    this.value = true;
-  }
-
-  handleBlur(e) {
-    if (DOM.activeElement !== this.element) {
+    if (e.type === 'focus') {
+      this.value = true;
+    } else if (DOM.activeElement !== this.element) {
       this.value = false;
     }
   }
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Ok, this is awesome. I didn't even know about this API option. @jdanyow Can you think of any reason not to do this? @bigopon What if we also removed the extra methods and just inlined the behavior in the switch statement?

bigopon:
Inlined. But I don't why the whole content got blown away and replaced, despite being the same ... Is that an issue ?

@EisenbergEffect The API also motivated me to propose /aurelia/binding/issues#604, where in each `Listener` binding, we use the instance itself as listener instead of a newly created arrow function. That proposal contains 2 changes, in which 1 is breaking (Allow multiples events attached on the same element, 1 in template declaration, and 1 in element consumption). In my opinion, maybe just pull in the changes for API usage first, and leave the other for future discussion.

EisenbergEffect:
@bigopon It looks like it could be some sort of whitespace issue.

@jdanyow Can you review this and let me know if you see any issues. This looks like a good change and something we might be able to leverage in other places as well.

bigopon:
@EisenbergEffect @jdanyow  I've also put in observer locator to address #273

jdanyow:
The API is new- it's from https://github.com/aurelia/binding/pull/620

Add `?w=1` to ignore whitespace differences on github. For example, here's this PR: https://github.com/aurelia/templating-resources/pull/291/files?w=1

bigopon:
I've reverted all the changes. Couldn't think of binding order when attempted to fix that

Issue #298, opened 2017-08-02
View model field bound to "element.ref" is null in "attached" method for a combination of "if.bind", "repeat.for" and animations


Alexey1:
**I'm submitting a bug report**

* **Library Version:**
1.4.0

**Please tell us about your environment:**
* **Operating System:**
Windows 8

* **Node Version:**
6.9.2

* **NPM Version:**
4.3.0

* **JSPM OR Webpack AND Version**
Aurelia CLI 0.30.1

* **Browser:**
Chrome 60

* **Language:**
TypeScript 2.4.2

**Current behavior:**
1. Create a project with au new.
2. Add aurelia-animator-velocity as described [here](https://github.com/aurelia/animator-velocity/issues/13#issuecomment-245799650).
3.
main.ts
```
import { Aurelia } from 'aurelia-framework'
import environment from './environment';

export function configure(aurelia: Aurelia) {
  aurelia.use
    .standardConfiguration()
    .plugin('aurelia-animator-velocity', instance => {
      instance.enterAnimation.options.duration = 1000;
      instance.leaveAnimation.options.duration = 1000;
    })
    .feature('resources');

  if (environment.debug) {
    aurelia.use.developmentLogging();
  }

  if (environment.testing) {
    aurelia.use.plugin('aurelia-testing');
  }

  aurelia.start().then(() => aurelia.setRoot());
}
```

app.ts
```
export class App {
  private toggle = true;
}
```

app.html
```
<template>
  <require from="./item-details"></require>
  <label>
  <input type="checkbox"
    checked.bind="toggle">
    toggle
  </label>
  <div>
    <div class="au-animate"
      if.bind="toggle">
      <div repeat.for="i of 3"
        class="au-animate">
        <item-details index.bind="i"></item-details>
      </div>
    </div>
  </div>
</template>
```

add item-details.html file in /src/ folder
```
<template>
  <div element.ref="detailsEl">${index}</div>
</template>
```

add item-details.ts file in /src/ folder
```
import { bindable } from "aurelia-framework";

export class ItemDetails {

  @bindable index: number;
  detailsEl: Element;

  attached() {
    console.log('attached', this.index, this.detailsEl);
  }

  detached() {
    console.log('detached', this.index);
  }
}
```

au run, load page, see "toggle" checkbox, list of three items. Click "toggle" checkbox twice under one second, see output in console:
```
[...]

Issue #300, opened 2017-08-03
repeat.spec.js tests repeat object from jspm package instead of one from the project source


Alexey1:
**I'm submitting a bug report**

* **Library Version:**
1.4.0

**Please tell us about your environment:**
* **Operating System:**
Windows 8

* **Node Version:**
6.9.2

* **NPM Version:**
4.3.0

* **JSPM OR Webpack AND Version**
Aurelia CLI 0.30.1

* **Browser:**
Chrome 60

* **Language:**
TypeScript 2.4.2

**Current behavior:**
repeat.spec.js tests file
jspm_packages/npm/aurelia-templating-resources@1.0.0-rc.1.0.2/repeat.js 
instead of file
src/repeat.js

STR:
1. Setup as described in Readme.md
2. Add "debugger;" line in one of the tests like so:
```
it('should do nothing if no items provided', () => {
      let didThrow = false;
      repeat.items = undefined;

      try {
        debugger;
        repeat.bind();
      } catch (e) {
        didThrow = true;
      }

      expect(didThrow).toBe(false);
    });
```
2. Run command ```karma start```

3. Open dev tools in karma browser

4. Refresh page if needed, wait until "debugger" line starts debug mode

5. Step into repeat.bind(); method

6. Notice the repeat.js file is from jspm package instead of the project source:
![image](https://user-images.githubusercontent.com/1244471/28925616-ff7265ea-786d-11e7-980f-4b3e86483481.png)

**Expected/desired behavior:**

Repeat object is loaded from src/repeat.js file.

**Additional observations**:
I guess the reason for this is that repeat object from Repeat class imported like this
```import {Repeat} from '../src/repeat';```
is never created. Instead one retrieved like this 
```component = StageComponent.withResources().inView('<div repeat.for="item of items"></div>').boundTo({ items: [] });```
```...```
```repeat = component.viewModel;```
is used and it comes from jspm package.

huochunpeng:
It seems this is left unattended for too long. I just created a duplicated issue, with some unsuccessful attempts to fix.

PR #303, opened 2017-08-21
fix(repeat): bug when "if.bind", "repeat.for" and animations are used in a certain way

Alexey1:
closes https://github.com/aurelia/animator-velocity/issues/21 and
https://github.com/aurelia/templating-resources/issues/298

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -103,7 +103,7 @@ export class Repeat extends AbstractRepeater {
     this.scope = null;
     this.items = null;
     this.matcherBinding = null;
-    this.viewSlot.removeAll(true);
+    this.viewSlot.removeAll(true, true);
     this._unsubscribeCollection();
   }
 


Alexey1:
Please, note, I didn't do according test change because existing tests container for ```repeat``` is set up in such way that it tests ```repeat``` from a jspm package instead of one from the source code. Reported in https://github.com/aurelia/templating-resources/issues/300 .

zewa666:
Thanks for the PR, I'll try to get as soon as possible to verify that the fix doesn't break the CSS Animator.

zewa666:
ok so I've checked the css animator with your fix and it works as well 👍 
since @jdanyow already review the change in one of the mentioned issues I think we can safely merge this one @EisenbergEffect

zewa666:
@Alexey1 thanks again for the help

Alexey1:
no problem

PR #308, opened 2017-09-18
Fix #307: Failing IfCustomAttribute tests

### LINKED ISSUE #307: "if" tests are broken
Tests use `sut = new If(viewFactory, viewSlot, taskQueue);` and later use `sut.valueChanged(false);` but the `valueChanged` method no longer exists:

https://github.com/aurelia/templating-resources/commit/8b0131aefb8f181fc110e036cae11589bbba1e37#diff-bbce44daacb259a7b129c4fa0d062406

cc @jods4

jods4:


jods4:
Reviewers may wonder why one test was removed, answer is here:
https://github.com/aurelia/templating-resources/issues/307#issuecomment-330378995

test/if.spec.js
```diff
@@ -134,13 +119,13 @@ describe('if', () => {
...
       delay(200).then(() => {          
-        sut.valueChanged(false);
+        sut.conditionChanged(false);
         return delay(400);
       }).then(() => {
-        sut.valueChanged(true);
+        sut.conditionChanged(true);
         return delay(600);          
       }).then(() => {  
         expect(viewSlot.children.length).toEqual(1);
```
jdanyow: @jods4 This assertion is failing for me. Does it work for you?

jods4: @jdanyow I did not look specifically but when I committed all `if` tests were (seemed?) green :(

I am away from code but if I should make a guess: I wonder if the problem is here:
https://github.com/aurelia/templating-resources/blob/master/src/if.js#L127

The thing that is broken in tests is that they call `conditionChanged` (`valueChanged` before) but they do not actually set `condition`, which can lead to discrepencies.

`conditionChanged` is an internal callback when `condition` changes, it's not meant to be used from outside.
What looks right to me would be `condition = false`  instead of `conditionChanged(false)`. For some reason I didn't have time to debug, this doesn't work at all even when flushing the micro-task queue just after. Not sure why, I thought it would work.

PR #318, opened 2017-10-06
refactor(Signal): use new signal mechanism

bigopon:
This PR follows & requires the new feature of value converter trigger in aurelia/binding#626

### DIFF
--- a/src/binding-signaler.js
+++ b/src/binding-signaler.js
@@ -1,16 +1,11 @@
-import {sourceContext} from 'aurelia-binding';
+import {sourceContext, signalBindings} from 'aurelia-binding';
+import * as LogManager from 'aurelia-logging';
 
 export class BindingSignaler {
-  signals = {};
 
   signal(name: string): void {
-    let bindings = this.signals[name];
-    if (!bindings) {
-      return;
-    }
-    let i = bindings.length;
-    while (i--) {
-      bindings[i].call(sourceContext);
-    }
+    LogManager.getLogger('binding-signaler')
+      .warn(`Binding signaler signal has been deprecated. Please use "import {signalBindings} from 'aurelia-binding' instead.`);
+    signalBindings(name);
   }
 }
--- a/src/binding-signaler.js
+++ b/src/binding-signaler.js
@@ -4,12 +4,11 @@ import * as LogManager from 'aurelia-logging';
 export class BindingSignaler {
 
   constructor() {
-    this.logger = LogManager.getLogger('binding-signaler');
+    LogManager.getLogger('binding-signaler')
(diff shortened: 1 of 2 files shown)


src/binding-signaler.js
```diff
@@ -1,16 +1,11 @@
...
-      return;
-    }
-    let i = bindings.length;
-    while (i--) {
-      bindings[i].call(sourceContext);
-    }
+    LogManager.getLogger('binding-signaler')
+      .warn(`Binding signaler signal has been deprecated. Please use "import {signalBindings} from 'aurelia-binding' instead.`);
+    signalBindings(name);
   }
```
jdanyow: What do you think about adding a constructor to BindingSignaler and putting the log message there?

bigopon: Sure, was just waiting for feedback so it's a bit ... lazy there 😆

src/binding-signaler.js
```diff
@@ -1,16 +1,15 @@
...
-    let bindings = this.signals[name];
-    if (!bindings) {
-      return;
-    }
-    let i = bindings.length;
-    while (i--) {
-      bindings[i].call(sourceContext);
-    }
+    this.logger
+      .warn(`Binding signaler signal has been deprecated. Please use "import {signalBindings} from 'aurelia-binding' instead.`);
```
jdanyow: I was thinking this would move to the constructor as well. This way the message appears only once, the first time the application instantiates the BindingSignaler.  Sorry I wasn't clear the first time.

bigopon:
@done 👍 
Really like the changes in signal binding behavior, looks a lot cleaner. But I think there is a concern: Signaling used to be synchronous, this change will make it asynchronous, what do you think ?

jdanyow:
That's a good point. Previously, when calling `signaler.signal('something')` the bindings would have been updated when the method returned. With these changes the bindings will update on the next flush of the micro task queue. This is great in terms of batching DOM updates, especially if multiple signals are being published but it's definitely a breaking change. I think this is a tiny breaking change, won't impact the core scenarios for signal (i18n, clock time, etc). Let's ship this as-is, with the following workaround available to restore the original behavior:

```js
import {TaskQueue} from 'aurelia-task-queue';
import {Container} from 'aurelia-dependency-injection';
import {singalBindings} from 'aurelia-binding';

// publish the signal... bindings will update on next task queue flush.
signalBindings('something');
// workaround: force the task queue to flush immediately.
Container.instance.get(TaskQueue).flushMicroTaskQueue();
// bindings are now updated...
```

If this turns out to be a big problem for folks we can do a quick release to bake-in the TaskQueue flush.

Issue #317, opened 2017-10-04
1.5.1 issue, nested if.bind click.delegate does not fire after parent container has been hidden and re-shown


discoaaron:
**I'm submitting a bug report**

* **Library Version:**
1.5.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
7.4.0

* **NPM Version:**
4.0.5

* **JSPM OR Webpack AND Version**
webpack 3.3.0

* **Browser:**
Chrome 61.0.3163.100
Firefox 56.0 (32-bit)

* **Language:**
TypeScript

**Current behaviour:**
Nested if.bind click delegate does not fire when parent container is hidden and re-shown

**Expected/desired behavior:**
nested if.bind click delegate should fire when parent container is re-shown

**Setup**
Below is a clone of the aurelia-skeleton repo with only changes to welcome.ts, welcome.html and package.json. _I was unable to use gist due to the need to change the areulia-templating dependancy, sorry!_
https://github.com/discoaaron/skeleton-typescript-webpack-issue-example

** Expected Behaviour steps **
npm install
npm start

navigate to http://localhost:8080/
Click test - alert shown
Click Toggle (div is hidden)
Click Toggle (div is shown)
Click test - alert shown (as expected)

**Erroneous behaviour steps**
Update aurelia-templating to latest
change 
"aurelia-templating-resources": "1.4.0"
to 
"aurelia-templating-resources": "^1.4.0"	

stop watch task (if still running)
npm update
npm start
	
Click test - alert shown
Click Toggle (div is hidden)
Click Toggle (div is shown)
Click test - alert not shown! (not expected, this should have displayed like normal)
	
I've done some debugging and this is what I have so far
 - the click.delegate doesn't get called
 - debugging the event-manager.js, handleDelegatedEvent function shows that target.delegatedCallbacks is null/ empty when it shouldn't be. 
 - I attempted to debug `if.js` and the newly created `if-core.js` but this was about the point where I'm unsure what I'm looking at.

jsobell:
This may be a bug, but remember that `if.bind` doesn't *hide* an element, it removes it from the DOM.
If you want to hide and show it, use `show.bind`. I think you'll find you find your code then works.

discoaaron:
Yeah, thanks!

[...]

Issue #315, opened 2017-10-03
Aurelia if custom attribute does not work if it's rebound


zenorbi:
**I'm submitting a bug report**

* **Library Version:**
1.5.1

**Please tell us about your environment:**
* **Operating System:**
OSX 10.13

* **Node Version:**
v8.6.0

* **NPM Version:**
5.3.0

* **JSPM OR Webpack AND Version**
aurelia-cli 0.31.3

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
Rebinding a view having an if custom-attribute does not rebind the subviews inside the if.

**Expected/desired behavior:**

* **What is the expected behavior?**
Here is a repo to reproduce the bug as the gist.run does not use the latest `aurelia-templating-resources`.
https://github.com/zenorbi/aurelia-if-bug
Here is the gist with a the older aurelia-templating-resources which works: https://gist.run/?id=ef8b6fb22488bafda3c15b975e9c1820

bigopon:
What browser did you get the wrong behavior ? It's working fine for me, the button is doing its job properly. Did you miss anything in bug reproduction instruction ?

zenorbi:
@bigopon Chrome Version 61.0.3163.100 (Official Build) (64-bit) also using Firefox 56.0 (64-bit)

Please make sure you use the repo I linked as the gist version uses an older `aurelia-templating-resources`.

bigopon:
Only happen to nested `if`. The `View` of nested `if` will not call `bind` for some reason

zenorbi:
Well not just in case of nested `if`s, but any `if` binding where you `unbind` then `bind` the view again which has an `if` attribute somewhere inside it. The outside `if` is just the simplest way to get a view then unbind and rebind it.
In our internal project we use a caching aurelia `ViewFactory` where the created view only has a single `if` binding but due to caching we reuse thus rebind the view multiple times.

nenadvicentic:
Same problem here.
After template is hidden and shown again, `delegate` binding on some items does not work any more.

```
<template if.bind="...">
    <button if.bind="..." click.delegate="doSomething()"></button>
</template>
```

jods4:
#317 is a duplicate of this but contains an analysis of the cause and a proposed solution in a comment.

@EisenbergEffect We can close this issue and track the problem with #317.

EisenbergEffect:
Closing as duplicate of an issue that we've merged the fix for. See #317

PR #324, opened 2017-10-11
fix(if-core.js): child view re-binding issue

discoaaron:
sets showing to false on unbind to ensure binding is carried out when view is re-shown
nested views were not re-bound when parent was re-shown

Resolves
https://github.com/aurelia/templating-resources/issues/317
https://github.com/aurelia/templating-resources/issues/315


mjwwit:
Maybe we should add some unit-test(s) for this bug to prevent this from happening again.

jods4:
@discoaaron Thanks for the fix!
Can you revert all the `dist` changes please? They are generated when publishing a new version.

jods4:
I had a look at my changes when adding `else` and that `showing = false` was not there before:
https://github.com/aurelia/templating-resources/commit/8b0131aefb8f181fc110e036cae11589bbba1e37#diff-bbce44daacb259a7b129c4fa0d062406L108

My analysis is that where the subtle change happened is in the behaviour of old `_runValueChanged` vs `_show`:
https://github.com/aurelia/templating-resources/commit/8b0131aefb8f181fc110e036cae11589bbba1e37#diff-bbce44daacb259a7b129c4fa0d062406L66

New `_show()` does nothing if `showing` whereas the old implementation would still create and bind a view, even if `showing == true`. 

That old code doesn't actually make much sense and the flag `showing` and the state of the component are not in sync -- yet it worked.

I like your new fix better, the flag `showing` should be reset to false when component is unbound, so that everything is consistent and code logical.

jods4:
@mjwwit agreed that a new unit test would be great. That old behavior was hard to get from code only. 
Care to submit one? 

I'm back from a break and have _lots_ of catch up to do. 😞

EisenbergEffect:
@mjwwit @discoaaron I'm happy to merge this pending the removal of dist. If we can add a unit test to cover this at the same time, even better. Otherwise, maybe a second PR to add the test for this case. We definitely want to prevent similar regressions in the future. Ping me when this is ready and I'll get out a release within the day.

mjwwit:
@EisenbergEffect I'll work on a separate PR the unit-test.

EisenbergEffect:
@jods4 Can you do a review and let me know if this is good now?

jods4:
@EisenbergEffect Good for me! :shipit:

discoaaron:
Hey guys!
Sorry for the delay I was trying to figure out how the tests work, unfortunately (after about an hour of trial and error) I was unable to create a test which isolated the problem. 

I'd really appreciate the help @mjwwit 👍

mjwwit:
@EisenbergEffect Actually, hold off on merging for a bit. I think I can send a PR to @discoaaron's fork. That way you can merge everything at once.

mjwwit:
@discoaaron @EisenbergEffect PR with unit-tests is ready: discoaaron/templating-resources#1
I've included unit-tests for most of `ifCore.unbind()`, since there was no coverage at all.

discoaaron:
Thanks a lot @mjwwit! I can confirm the new tests exploit the bug.
Over to you @EisenbergEffect or @jods4

huochunpeng:
Too sad, this fix likely introduced a new issue #328.
I have to go back to v1.4.0 for old battle tested if.bind.

Issue #322, opened 2017-10-09
Else CustomAttrbiute not having bindingContext set


fragsalat:
**I'm submitting a bug report**

* **Library Version:**
1.5.1

**Please tell us about your environment:**
* **Operating System:**
Linux (Ubuntu 17.10)

* **Node Version:**
8.4.0

* **NPM Version:**
5.4.2

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
The element having the else attribute has no binding context which leads to missing data when the initial condition is false.
Example:
https://gist.run/?id=371fa4bf53068f0dd4463684b165cb66

**Cause**
The cause for this problem might be that the `If.conditionChanged()` function is called when the if is bound. Since the if attribute appears mostly before the else attribute, the else attribute has no bindingContext yet.
https://github.com/aurelia/templating-resources/blob/master/src/if.js#L23

**Possible solution**
Call the conditionChanged function also when else is bound or just use the binding context of the if attribute.

fragsalat:
Do you need support with this? I mean I can fix this but do you have time to review, merge, test and release the code?

jdanyow:
@fragsalat we'd definitely take a PR for this

huochunpeng:
I got exactly the same issue just now when I try the new `else` binding :-(
Just composed an almost identical gist-run.

Having additional binding inside a branch is a common use case. The else binding is a neat new feature, but it looks like it needs more test coverage before I can trust it in production.

fragsalat:
@huochunpeng I fixed this now and it just has to get merged and released :)

jods4:
This should be fixed in the current release, I think we can close this issue? 
@EisenbergEffect

PR #325, opened 2017-10-11
fix(if-else): Fixed missing binding context when rendering else block

fragsalat:
Issue: #322 

This PR fixes the issue that an else block is rendered without a bindingContext in case that the condition is falsy and the view is rendered initially.
I also added some simple unit tests to check if the correct block is rendered and the swapping logic works properly.

### DIFF
--- a/src/else.js
+++ b/src/else.js
@@ -13,11 +13,6 @@ export class Else extends IfCore {
 
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
-
-    if (!this.ifVm) {
-      throw new Error('There must be a node with an if attribute before the else attribute');
-    }
-
     this.ifVm.conditionChanged(this.ifVm.condition);
   }
 
@@ -35,6 +30,6 @@ export class Else extends IfCore {
       throw new Error("Can't find matching If for Else custom attribute.");
     }
     this.ifVm = previous.au.if.viewModel;
-    this.ifVm.else = this;
+    this.ifVm.elseVm = this;
   }
 }
--- a/src/else.js
+++ b/src/else.js
@@ -13,7 +13,10 @@ export class Else extends IfCore {
 
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
-    this.ifVm.conditionChanged(this.ifVm.condition);
+    // Render on initial
+    if (!this.ifVm.condition) {
+      this._show();
+    }
   }
 
   _registerInIf() {
--- a/src/else.js
+++ b/src/else.js
@@ -11,6 +11,16 @@ export class Else extends IfCore {
     this._registerInIf();
   }
 
+  bind(bindingContext, overrideContext) {
+    super.bind(bindingContext, overrideContext);
+
+    if (!this.ifVm) {
+      throw new Error('There must be a node with an if attribute before the else attribute');
+    }
+
+    this.ifVm.conditionChanged(this.ifVm.condition);
+  }
+
   _registerInIf() {
     // We support the pattern <div if.bind="x"></div><div else></div>.
     // Obvisouly between the two, we must accepts text (spaces) and comments.
@@ -24,7 +34,7 @@ export class Else extends IfCore {
     if (!previous || !previous.au.if) {
       throw new Error("Can't find matching If for Else custom attribute.");
     }
-    let ifVm = previous.au.if.viewModel;
-    ifVm.else = this;
+    this.ifVm = previous.au.if.viewModel;
+    this.ifVm.else = this;
   }
 }
--- a/src/if.js
+++ b/src/if.js
@@ -20,7 +20,10 @@ export class If extends IfCore {
   */
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
-    this.conditionChanged(this.condition);
+    // Else attribute will trigger change when it's bound
+    if (!this.else || this.else.isBound) {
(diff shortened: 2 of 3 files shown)


huochunpeng:
Thanks for the fast fix.
You forgot to replace `fit(...)` with `it(...)`...

fragsalat:
Oh my god. Sry my mistake -> it's removed now

fragsalat:
Any update?

jods4:
Sorry I was away for a while.
I did the `else` feature and I'll have a look at the PR.
The additional test is especially useful.

src/if.js
```diff
@@ -20,7 +20,10 @@ export class If extends IfCore {
   */
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
-    this.conditionChanged(this.condition);
+    // Else attribute will trigger change when it's bound
+    if (!this.else || this.else.isBound) {
```
jods4: Let me recap the problem: 
When `if` is bound, if `condition` is `false` then `else._show()` is called but at this point `else.bind()` hasn't happen so everything crash and burns.

Re-thinking what is supposed to happen during bind, can I suggest a more straightforward solution:
```js
// inside If::bind()
if (this.condition) this._show();

// inside Else::bind()
if (!this.if.condition) this._show();
```

Would that work?
It seems simpler to understand.

src/else.js
```diff
@@ -24,7 +34,7 @@ export class Else extends IfCore {
     if (!previous || !previous.au.if) {
       throw new Error("Can't find matching If for Else custom attribute.");
     }
-    let ifVm = previous.au.if.viewModel;
-    ifVm.else = this;
+    this.ifVm = previous.au.if.viewModel;
```
jods4: The equivalent link to Else viewmodel in If is simply called `else`. For consistency, can we rename `ifVm` to `if` -- or rename `else` to `elseVm`?

EisenbergEffect:
Ping me when this is ready and I'll get it out asap.

src/else.js
```diff
@@ -11,6 +11,16 @@ export class Else extends IfCore {
     this._registerInIf();
   }
 
+  bind(bindingContext, overrideContext) {
+    super.bind(bindingContext, overrideContext);
+
+    if (!this.ifVm) {
```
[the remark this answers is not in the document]
fragsalat: @jods4 I was thinking about if the else is specified before the if so that it could be in the context but bound before. But after you said it, I recognized that this might not be possible since the if is registered in the constructor.

src/else.js
fragsalat: +1 for elseVm for better understanding

src/if.js
fragsalat: @jods4 I'm not quite sure if this is possible since the else.bind function won't be called when If's condition has changed. I guess this is why you implemented it that way^^

fragsalat:
@jods4 I removed the obsolete error and fixed the typo

src/if.js
jods4: > the else.bind function won't be called when If's condition has changed.

It is called when the component is first bound. It is not meant to be called when the condition later changes.

`bind()` functions are called once at startup for both `if` and `else`, they ensure the template is visible when it is first bound.

If the condition later changes, `If::conditionChanged()` is called and updates both element properly.

Am I missing something?

fragsalat:
@jods4 Ping

jods4:
@fragsalat Pong
What about my last comment https://github.com/aurelia/templating-resources/pull/325#discussion_r145387032 ?

src/if.js
fragsalat: But in this case only the initial render would work or?
If the if condition get's bound to new values the bind function should be evaluated for the if only.

Actually I'm not really sure here if your way would work in all scenarios but this one does and it has not really an impact to change it or I'm wrong here? :)

jods4: If what I proposed above works (I think it should), it's much simpler and easier to understand so I'd like to include it.
Simpler code is very important for people who will maintain or contribute to Aurelia in the future -- or just because it means that the risk of bug is lower.

In fact I had to read your code several time to figure out how everything clicks together. Conditions like `this.elseVm.isBound` in `If::bind()` really make no sense at first glance -- heck thinking about it again: why call `conditionChanged` in `If::bind()` when `else` was bound, as `else` would actually have called `if.conditionChanged` in its own bind?!

RE your questions: when `condition` changes, `bind` is not called again. Changing the `bind()` implementation doesn't impact that, it only matters when the controller is first connected by Aurelia.

EisenbergEffect:
@fragsalat I've been staring at your avatar forever, and it just dawned on me that it's Natsu. Nice.

fragsalat:
@jods4 I checked it and u were right :) It's changed now.
@EisenbergEffect Yeah u're right :) Didn't expected that u know Natsu ^^

EisenbergEffect:
@fragsalat My wife and I have watched every episode of Fairy Tail. It's one of our favorite anime. Eagerly waiting for the next season. Hopefully 2018 will deliver 😸

fragsalat:
Wow. You're a lucky guy that u're wife is watching this with you. Mine is annoyed from the female (Japanese) voices and doesn't understand the story -,-
From my early days my favorite is still one piece but since some years I love Fairy Tail as well :)

EisenbergEffect:
Thanks @jods4 I'll get it in and hopefully released tonight.

fragsalat:
:+1:  Thx to u guys :)

PR #330, opened 2017-10-25
fix(if): fix nested if bindings

### LINKED ISSUE #328: Leaking (the new) if binding introduced in v1.5.2

* **Library Version:**
1.5.2

* **Operating System:**
OSX 10.x

* **Node Version:**
8.5.0

* **NPM Version:**
5.5.1

CLI

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
https://github.com/huochunpeng/ifff

A leaking if binding behind nested if binding

Run the test app, just click next and next, all conditional DOM are not removed.

Pretty sure this is introduced by #324 which attempt to fix another issue on the new if/else implementation.

This issue doesn't exist on v1.5.1 before #324 fix.

bring back old if.bind please...
The new if/else doesn't justify the pain...

mjwwit:
- Revert (flawed) fix from #324
- Force rebind of view (if not bound) if the if-binding is rebound and condition is `true`
- Fix unit-tests for repeat, which were failing because of work done in #303

Closes #328

### DIFF
--- a/src/if-core.js
+++ b/src/if-core.js
@@ -31,7 +31,6 @@ export class IfCore {
     // For example a view could be returned to the cache and reused while it's still
     // attached to the DOM and animated.
     if (!this.viewFactory.isCaching) {
-	    this.showing = false;
       return;
     }
 
@@ -45,8 +44,8 @@ export class IfCore {
     this.view = null;
   }
 
-  _show() {
-    if (this.showing) {
+  _show(force) {
+    if (this.showing && !force) {
       return;
     }
 
--- a/src/if.js
+++ b/src/if.js
@@ -21,7 +21,7 @@ export class If extends IfCore {
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
     if (this.condition) {
-      this._show();
+      this._show(true);
     }
   }
 
--- a/test/if.spec.js
+++ b/test/if.spec.js
@@ -63,25 +63,6 @@ describe('if', () => {
     expect(sut.view).toBeNull();
   });
 
-  it('should correctly unbind view without removing it when viewFactory is caching and being unbound itself', () => {
-    let view = new ViewMock();
-    sut.view = view;
-    sut.showing = true;
-    viewFactory.isCaching = false;
-    spyOn(viewSlot, 'remove');
-    spyOn(view, 'unbind');
-    spyOn(view, 'returnToCache');
-
-    sut.unbind();
-    taskQueue.flushMicroTaskQueue();
-
-    expect(sut.showing).toBeFalsy();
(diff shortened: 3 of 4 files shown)


jods4:
Sorry I didn't see this and I was working on a fix on my own yesterday.

I had a glance at your code and I'm unsure how you actually fixed #328? 
What in your code hides a nested if that was unbound in showing state and binds back in not showing state?

If you don't mind I would rather close this PR and merge #330.

I see you created an additional test, it would be nice to merge it, though.

mjwwit:
> I had a glance at your code and I'm unsure how you actually fixed #328?
What in your code hides a nested if that was unbound in showing state and binds back in not showing state?

The incorrect showing of the nested `if` was actually due to #324, I fixed the issue by reverting that change.
Fixing the actual issue (#315 and #317) was done by forcing an `if` binding to re-bind its child view if its being bound itself.
The issue was that this logic was skipped because of the `showing` boolean flag in the nested `if` view-model. This view-model keeps its state even if unbound. When unbound though, the change callbacks do not trigger and it doesn't get updated. When the `if` then gets rebound, `_show` is called but the `showing` flag is still `true`, causing the child view binding to be skipped.

I still want to make more of a high-level unit test actually checking for the correct behavior, but I'm unable to get the `ComponentTester` from aurelia-testing working with the `if` `customAttribute`.

> If you don't mind I would rather close this PR and merge #330.
This is #330. I'm confused...

mjwwit:
I now see you meant #331. Our fixes are very similar. There is a difference though, after some testing I chose not to call `_hide` in `bind()` when the `condition` is falsy. This caused errors in some cases. Also, I found that hiding wasn't the problem, the problem was re-binding the child-view if the `condition` is truthy.

Hilarious that you ran into issues with `ComponentTester` as well. I think it's a circular dependency issue (aurelia-testing -> aurelia-bootstrapper -> aurelia-templating-resources). When it tries to register the resources you get an error saying it already has an `if` attribute registered. I'll see if I can get around that somehow.

As to which PR gets merged, I really don't care, as long as the issue gets fixed.

mjwwit:
After close inspection your PR seems more complete. I forgot to add the fix to `else` and I'm executing more (redundant) logic when a nested if is bound. Your PR simply re-binds the child-view and that's it.

I'll update this PR to add the much desired nested `if` tests when I get them working.

mjwwit:
Actually, I'll send a PR to your fork, that way I don't have to wait for your PR to be merged.

jods4:
Thanks, I merged your changes and additional tests!

I had to disable the nested ones though, as they didn't seem to work.
Honestly, I doubt `ComponentTester` can work in that karma/unit testing context. :(

Issue #327, opened 2017-10-19
if.bind both in and outside the custom element causes ref not to be populated when re-binding


aqm5181:
**I'm submitting a bug report**

* **Library Version:**
1.5.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10 x64

* **Node Version:**
8.6.0

* **NPM Version:**
5.3.0

* **JSPM OR Webpack AND Version**
Aurelia CLI 0.31.3

* **Browser:**
Chrome 61.0.3163.100

* **Language:**
TypeScript 2.08

**Current behavior:**

1. Create project with au new
2. 

main.ts
```
import {Aurelia} from 'aurelia-framework'
import environment from './environment';

export function configure(aurelia: Aurelia) {
  aurelia.use
    .standardConfiguration()
    .feature('resources');

  if (environment.debug) {
    aurelia.use.developmentLogging();
  }

  if (environment.testing) {
    aurelia.use.plugin('aurelia-testing');
  }

  aurelia.start().then(() => aurelia.setRoot());
}
```

app.ts
```
export class App {
  public toggle = false;
  public elemvisible = true;
}
```

app.html
```
<template>
  <require from="./item-details"></require>
  <label>
    <input type="checkbox"
           checked.bind="elemvisible">
    Elem Visible
    <input type="checkbox"
           checked.bind="toggle">
    toggle
  </label>
  <div>
    <div if.bind="toggle">
        <item-details elemvisible.bind="elemvisible"></item-details>
    </div>
  </div>
</template>
```

item-details.ts
```
import { bindable } from "aurelia-framework";
import {bindingMode} from "aurelia-binding";

export class ItemDetails {

  @bindable elemvisible: boolean;
  detailsEl: Element;

  attached() {
    console.log('attached', this.detailsEl);
  }

  detached() {
    console.log('detached', this.detailsEl);
  }
}
```

item-details.html
```
<template>
  <div if.bind="elemvisible">
    Stuff:
    <div element.ref="detailsEl">Displayed</div>
  </div>
</template>
```

When toggling the "toggle" checkbox the first time, elDetails in custom element is referenced correctly, however subsequent toggles return null for it. This only happens if there is an if.bind both within the custom element and also in the parent page.

Example: https://gist.run/?id=a981dc20a5cc72439349c02df465ebe7

Similar to issue #298 except for the additional if.bind in the parent page.

**Expected/desired behavior:**

* **What is the expected behavior?**
[...]

Issue #326, opened 2017-10-12
[BUG] ref is null in attached callback in certain nested if.bind


huochunpeng:
**I'm submitting a bug report**

* **Library Version:**
1.5.1

**Please tell us about your environment:**
* **Operating System:**
OSX 10.x

* **Node Version:**
8.5.0

* **NPM Version:**
8.5.0

* **JSPM OR Webpack AND Version**
aurelia-cli 0.31.3

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
A parent component with an inner child component, both behind an if bind.

The app
```
<template>
  <require from="./test-container"></require>
  
  <test-container if.bind="show"></test-container>
  
  <button click.delegate="show = !show">toggle</button>
</template>
```

The test-container component
```
<template>
  <require from="./child"></require>
  <div if.bind="true">
    <child></child>
  </div>
</template>
```

The child component
```
<template>
  <p ref="el">lorem</p>
  <p>message: ${message}</p>
</template>
```

In child component js file.
```
export class Child {
  attached() {
    const m = 'el: ' + this.el;
    console.log(m);
    
    this.message = m;
    console.log('message: ' + this.message);
  }
}
```

The `this.el` has correct value when first time the parent component is attached, but not second time.

Try "toggle" button few times, you can see in browser console, it prints
```
el: null
message: el: null
```

This issue only happens on latest aurelia version, I suspect it's introduced by new `if` binding implementation.

**Expected/desired behavior:**
https://gist.run/?id=24bddb1ca8e502bcac280fbb3368dd30

You have to see the problem in browser console.

The most bizarre part is `message` somehow shows correctly as "el: [object HTMLParagraphElement]" in the DOM, but it clearly reads "el: null" in console!

* **What is the expected behavior?**
Bring my ref back pls.

* **What is the motivation / use case for changing the behavior?**
This is a regression bug to be fixed.

huochunpeng:
I confirm there is no issue on old v1.4.0.

@jdanyow @jods4 can we revert the commit of new if/else implementation? Even if you fix this bug in time, the change introduced by the new implementation is too significant, which worries me about unintended edge cases.

It costed hell of my time on debugging. I have to fix aurelia-templating-resources version to v1.4.0 in my package.json to avoid futher pain.

discoaaron:
Hello @huochunpeng

I've submitted a PR with a proposed fix for this issue, feel free to test this solution on your application. 
[...]

PR #331, opened 2017-10-25
fix(if): nested ifs don't work properly

### LINKED ISSUE #328: Leaking (the new) if binding introduced in v1.5.2

jods4:
fixes #328, #327, #326, #322, #317, #315

### DIFF
--- a/src/else.js
+++ b/src/else.js
@@ -14,7 +14,10 @@ export class Else extends IfCore {
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
     // Render on initial
-    if (!this.ifVm.condition) {
+    if (this.ifVm.condition) {
+      this._hide();
+    }
+    else {
       this._show();
     }
   }
--- a/src/else.js
+++ b/src/else.js
@@ -16,8 +16,7 @@ export class Else extends IfCore {
     // Render on initial
     if (this.ifVm.condition) {
       this._hide();
-    }
-    else {
+    } else {
       this._show();
     }
   }
--- a/src/if-core.js
+++ b/src/if-core.js
@@ -46,12 +46,12 @@ export class IfCore {
 
   _show() {
     if (this.showing) {
-      // Ensures the view is bound. 
+      // Ensures the view is bound.
       // It might not be the case when the if was unbound but not detached, then rebound.
       // Typical case where this happens is nested ifs
       if (!this.view.isBound) {
         this.view.bind(this.bindingContext, this.overrideContext);
-      }      
+      }
       return;
     }
 
--- a/src/if-core.js
+++ b/src/if-core.js
@@ -31,7 +31,6 @@ export class IfCore {
     // For example a view could be returned to the cache and reused while it's still
     // attached to the DOM and animated.
     if (!this.viewFactory.isCaching) {
-	    this.showing = false;
       return;
     }
 
@@ -47,6 +46,12 @@ export class IfCore {
 
   _show() {
     if (this.showing) {
+      // Ensures the view is bound. 
+      // It might not be the case when the if was unbound but not detached, then rebound.
+      // Typical case where this happens is nested ifs
+      if (!this.view.isBound) {
+        this.view.bind(this.bindingContext, this.overrideContext);
+      }      
       return;
     }
 
--- a/src/if.js
+++ b/src/if.js
@@ -23,6 +23,9 @@ export class If extends IfCore {
     if (this.condition) {
       this._show();
     }
+    else {
+      this._hide();
+    }
   }
 
   /**
--- a/src/if.js
+++ b/src/if.js
@@ -22,8 +22,7 @@ export class If extends IfCore {
     super.bind(bindingContext, overrideContext);
     if (this.condition) {
       this._show();
-    }
-    else {
+    } else {
       this._hide();
     }
   }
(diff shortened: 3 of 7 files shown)


jods4:
This should fix all reported issues with the `if` refactoring. 

I have coded two short repro of reported bugs to confirm it's fixed and to prevent future regressions, unfortunately after more than 6 hours of trying I can't get `ComponentTester` to work in karma tests.
I would really like to commit those or something equivalent, if anyone can help.

Here are the two non-comitted tests:
```js
  describe('(nested)', () => {
    let taskQueue, component, model;

    beforeEach(() => {
      model = { a: true, b: true, clicked() {} };
      component = new ComponentTester();
      component.bootstrap(aurelia => {
        aurelia.use.standardConfiguration();
        taskQueue = aurelia.container.get(TaskQueue);
      });
      component
        .inView(`<div if.bind="a" id="a">
                   <div if.bind="b" id="b" click.delegate="clicked()">
                   </div>
                 </div>`)
        .boundTo(model);        
    });

    afterEach(() => component.dispose());

    it('should rebind even if the view is visible; issue #317', done => {
      component.create(bootstrap)
               .then(_ => {
                 let spy = spyOn(model, "clicked");
                 model.a = false;
                 taskQueue.flushMicroTaskQueue();
                 model.a = true;
                 taskQueue.flushMicroTaskQueue();
                 document.getElementById("b").click();
                 
                 expect(spy).toHaveBeenCalled();
                 done();
               });
    });

    it('should update view when rebound; issue #328', done => {
      component.create(bootstrap)
               .then(_ => {                
                 model.a = false;
                 taskQueue.flushMicroTaskQueue();
                 model.b = false;
                 model.a = true;
                 taskQueue.flushMicroTaskQueue();
 
                 expect(document.getElementById("a")).not.toBeNull();
                 expect(document.getElementById("b")).toBeNull();
                 done();
               });
    });
  });
```

src/if-core.js
```diff
@@ -47,6 +46,12 @@ export class IfCore {
 
   _show() {
     if (this.showing) {
+      // Ensures the view is bound. 
+      // It might not be the case when the if was unbound but not detached, then rebound.
+      // Typical case where this happens is nested ifs
+      if (!this.view.isBound) {
+        this.view.bind(this.bindingContext, this.overrideContext);
+      }
```
mjwwit: More trailing spaces.

src/if-core.js
```diff
@@ -47,6 +46,12 @@ export class IfCore {
 
   _show() {
     if (this.showing) {
+      // Ensures the view is bound.
```
mjwwit: There's a trailing space here, causing a linter error.

src/if.js
```diff
@@ -23,6 +23,9 @@ export class If extends IfCore {
     if (this.condition) {
       this._show();
     }
+    else {
```
mjwwit: Same brace issue as with `src/else.js`.

src/else.js
```diff
@@ -14,7 +14,10 @@ export class Else extends IfCore {
   bind(bindingContext, overrideContext) {
     super.bind(bindingContext, overrideContext);
     // Render on initial
-    if (!this.ifVm.condition) {
+    if (this.ifVm.condition) {
+      this._hide();
+    }
+    else {
```
mjwwit: According to ESLint, the `else` should be on the same line as the closing brace from the `if`.

mjwwit:
@jods4 I've sent you a PR (jods4/templating-resources#1) with the unit-test from #330 (and additional tests), linter fixes, and actual integration tests for nested if bindings. This was a real b*tch to get working btw.

jods4:
@mjwwit Thanks, it's merged! Although I had to disable the `ComponentTester` tests. :(

jods4:
In addition to my own tests, we have user confirmation in #328 that the fix works for him.
@EisenbergEffect I think we can publish this?

mjwwit:
@jods4 That sucks, I spent tons of time figuring a way around the duplicate attribute conflict... It worked for me using Karma and Chrome 61 on macOS... What issue are you running into?

jods4:
@mjwwit I spent ton of time on that as well... which is why I gave up in the end.
I've tested in Firefox 57.0b11 on Windows 10.

A few observations I made when trying to get `ComponentTester` to work:
- It uses JSPM, whose `config.json` file locks to very old libs. When I updated everything `jspm update` I got lots of new tests failures.
- JSPM seems to -- at least sometimes if not always -- load `aurelia-templating-resources` from NPM rather than from `dist`, which is the system under test.
- `ComponentTester` only seems to work for the first test that uses it and fail for subsequent ones. So using `fit` (focused test) on any of those actually worked but never when running all of them.

mjwwit:
Alright. I'll see if I can migrate this project to the new TypeScript build soon, maybe that'll fix these problems and finally allow for proper integration tests.

EisenbergEffect:
@mjwwit This is a larger, more complicated, more risky project to move to TS. I'd prefer to get a few more smaller ones done first ;)

nenadvicentic:
@EisenbergEffect Where are you tracking porting to Typescript?

PR #334, opened 2017-11-29
Added 'else' to lifecycleOptionalBehaviors in analyze-view-factory

bulldetektor:
The 'else' attribute was added to aurelia-templating-resources 1.5.0, but seems like it was left out from the 'livecycleOptionBehaviors'. This leads to unnecessary work when removing av view from the DOM that uses the 'else' attribute.

### DIFF
--- a/src/analyze-view-factory.js
+++ b/src/analyze-view-factory.js
@@ -2,7 +2,7 @@
 * Behaviors that do not require the composition lifecycle callbacks when replacing
 * their binding context.
 */
-export const lifecycleOptionalBehaviors = ['focus', 'if', 'repeat', 'show', 'with'];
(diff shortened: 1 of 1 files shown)


bulldetektor:
Note; there was a failing Karma test;

> Chrome 62.0.3202 (Windows 10 0.0.0) repeat unbind should remove all views FAILED
>         Expected spy removeAll to have been called with [ true, true ] but actual calls were [ true ].
>             at UserContext.eval (test/repeat.spec.js:90:34)

This test failed prior to the changes I made and still fails afterwards.

gheoan:
Should `hide` also be added to this list?

jdanyow:
thanks!

PR #336, opened 2017-12-24
fix(debounce): debounce binding call instead updatesource

### LINKED ISSUE #319: [Proposal]: debounce fix
Currently debounce is working incorrectly because of [this line in binding-expression](https://github.com/aurelia/binding/blob/master/src/binding-expression.js#L68)

```js
// line 68
if (newValue !== this.sourceExpression.evaluate(this.source, this.lookupFunctions)) {
  this.updateSource(newValue);
}
```

Reverting the value quickly will cause the new `updateSource` to not be called.

Proposal fix: Debounce binding behavior will alter the binding even further, when `methodToDebounce` is `updateSource`:

  * Modify `call`: check if it's `targetContext`, we call `updateSource` directly and let the diff happen there. This works but is bad because `targetContext` is private constant of aurelia binding.
  * Modify `updateSource` with another version that has value diffing.
  * [Demo of fix](https://gist.run/?id=85c510c2c33388258fa08bf1f41f16b5) for aurelia/binding#625 & ~~aurelia/binding#485~~

@jdanyow @EisenbergEffect @salchi

bigopon:
I tried to fix the throttle but pretty much gave up on the test. I think we should fix debounce first, regardless throttle status.

resolves #319
fixes aurelia/binding#625
fixes aurelia/binding#485

The gist of throttle working is here https://gist.run/?id=4fec87bd40bdfbd7ed10e7ad8f37fdd8 But I couldn't apply jasmin clock against it.

@jdanyow

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,5 +1,12 @@
 System.config({
   defaultJSExtensions: true,
+  transpiler: "babel",
+  babelOptions: {
+    "optional": [
+      "runtime",
+      "optimisation.modules.system"
+    ]
+  },
   paths: {
     "github:*": "jspm_packages/github/*",
     "aurelia-templating-resources/*": "dist/*",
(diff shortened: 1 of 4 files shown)


huochunpeng:
What's holding back this PR?

PR #340, opened 2018-02-05
perf(all): synchronous resources registration

bigopon:
@EisenbergEffect this is to go together with https://github.com/aurelia/framework/pull/858 

I have delayed the load of view factory & compiling process in the framework configuration until all resources have been loaded. This changes isn't necessary to avoid breaking existing code, it's just a mean to use the new api.


EisenbergEffect:
@bigopon Is this an old PR?

bigopon:
Yes, but it will still works fine after new framework PR, as it still use same API `globalResources` @EisenbergEffect

EisenbergEffect:
Gotcha!

bigopon:
@EisenbergEffect This is in good state to merge after aurelia/framework#885

EisenbergEffect:
Will final review this and the other PR this weekend. Thanks!

EisenbergEffect:
@bigopon Can you remove the `d.ts` file from this PR? After that, I'm ready to merge.

bigopon:
I think we should also move `@inject` to `static inject` for DOM.Element injection. These framework code folks cannot control and will have hard to if PLATFORM code is not called at appropriate time @EisenbergEffect

EisenbergEffect:
@bigopon I saw you do that in one place already...I think it was for the SlotAttribute. It makes sense. If you want to add that to this PR, that's fine with me.

EisenbergEffect:
@bigopon It's in! What's next? Are there any more PRs besides the templating-router one that we need to get in to complete the feature?

bigopon:
That's it, `templating-router` is the last one for first batch @EisenbergEffect

PR #343, opened 2018-03-17
fix(UpdateTrigger): sync with binding

bigopon:
- Ensure we dispose the listener before resetting the handler back to its original version.
- Use `EventSubscriber`

@EisenbergEffect this requires `aurelia-binding@1.7.x` and the fix at aurelia/binding#674

### DIFF
--- a/src/update-trigger-binding-behavior.js
+++ b/src/update-trigger-binding-behavior.js
@@ -1,14 +1,9 @@
-import { bindingMode, EventManager } from 'aurelia-binding';
+import { bindingMode, EventSubscriber } from 'aurelia-binding';
 
 const eventNamesRequired = 'The updateTrigger binding behavior requires at least one event name argument: eg <input value.bind="firstName & updateTrigger:\'blur\'">';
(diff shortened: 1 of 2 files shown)


src/update-trigger-binding-behavior.js
```diff
@@ -29,13 +24,15 @@ export class UpdateTriggerBindingBehavior {
...
-    let handler = this.eventManager.createElementHandler(events);
+    let handler = new EventSubscriber(events);
     targetObserver.handler = handler;
   }
 
   unbind(binding, source) {
     // restore the state of the binding.
+    binding.targetObserver.handler.dispose();
     binding.targetObserver.handler = binding.targetObserver.originalHandler;
+    // console.log(binding.targetObserver.handler);
```
StrahilKazlachev: Leftover from testing?

bigopon: I even had `fit(...` not cleaned up. Thanks!

bigopon:
In the matter of lib compat, this could be a minor breaking changes. If we have any other using private api `EventManager.createElementHandler`, we can add it back with a deprecation note.

PR #344, opened 2018-03-27
feat(css-resource): add configuration option to remove injected styles

fkleuver:
This feature is requested now and then, and I'd also like to have it myself. It's probably not the cleanest solution, but I've tested it locally with webpack and works fine.

It's opt-in and will not change anything for existing applications.

Usage requires manually configuring aurelia and passing in the option, like so:

```ts
au.use
  .defaultBindingLanguage()
  .history()
  .router()
  .eventAggregator();
au.use.plugin(PLATFORM.moduleName('aurelia-templating-resources'), (opts: any) => {
  opts.removeInjectedStylesOnBeforeUnbind = true;
});
```

What this option will do is register the ViewEngineHooks (which normally only happens for `as="scoped"`), keep track of the injected StyleNode and remove it on `beforeUnbind`.

Since removing it will actually keep it removed across re-visits, it will add it again on `beforeBind` in case it hasn't been added yet by `beforeCompile`.

I also spotted a bug in the existing code: `} else if (this._global && !this.owner._alreadyGloballyInjected)`. It's possible that this prevented `as="scoped"` from working correctly. I changed `this._global` to `this.owner._global` which should fix that as well.

### DIFF
--- a/doc/article/en-US/remove-injected-styles.md
+++ b/doc/article/en-US/remove-injected-styles.md
@@ -1,0 +1,21 @@
+## Removing custom element injected styles
+
+When you have a style resource associated with a custom element, a style node containing the CSS is injected into the document's `<head>` node. Normally that style node will stay there, even after the custom element is detached from the DOM. Browser support for `as-scoped` and `ShadowDOM` is still thin and thus there is no easy way to remove injected styles again.
+
+[This PR](https://github.com/aurelia/templating-resources/pull/344) is an attempt to address that. The behavior needs to be turned on explicitly in your `main` file where aurelia is configured. This is done by replacing  `au.use.standardConfiguration()` with the following explicit configuration:
+
+```ts
+au.use
+  .defaultBindingLanguage()
+  .history()
+  .router()
+  .eventAggregator()
+  // This also replaces the .standardResources() call
+  .plugin(PLATFORM.moduleName('aurelia-templating-resources'), opts => {
+    opts.removeInjectedStylesOnBeforeUnbind = true;
+  });
+```
+
+This will enable the behavior globally. Per-element (or other context) configuration may be considered if there seems to be a demand.
+
+Please note that this should be considered an experimental feature and needs thorough testing before going in production. There may also be performance implications. Should you encounter any issues, feel free to leave a comment [here](https://github.com/aurelia/templating-resources/pull/344).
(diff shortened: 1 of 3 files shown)


Alexander-Taran:
package.lock.json intentional?

fkleuver:
Most certainly isn't, fixed it. Thanks :)

bigopon:
Will this affects elements with multiple instances ? Unbinding one of them will cause the style to be removed for the rest ?

fkleuver:
Elements will only track the nodes they added themselves, so if 10 of the same are rendered then the 1st element has added the style; only if *that* element is unbound, the style will be removed ("for all of them", yes). The other 9 have no influence in this regard.

So arguably you'd want to keep track of all instances of an element, and only remove the style if **all** of them are gone. It's possible but a lot of work, and probably not worth it. It's probably worth a mention in the docs though.

Alexander-Taran:
Ouch.. exactly what I was asking you..
So basically I can't opt into it.
Because I <require> syles for every custom element
so any repeater would brake styling of the repeated elements on removal of 1st element?

If merged - it should be not mentioned anywhwere.
No instructions whatsoever on that there is a possibility like this.
No hints even.

It will be misused. And will only lead to issues and feature requests.
Nah.. nobody will see it as a feature request.. Straight out bug report that'd be (-:

EisenbergEffect:
I missed that detail. I don't think we want to merge something that would remove styles and break components. A simple instance counter might make it work...

fkleuver:
Forgive me, I was a bit tired when I wrote that. You're right this is silly :)

I'll see if I can fix this properly.

Alexander-Taran:
@fkleuver nothing to apologize for. That's what code review is for. (-:

fkleuver:
The more I look into this the less I understand how this somehow appeared to work in a small project of mine. Probably something to do with the lazy loading of routes.

In webpack, all imported custom elements (either via `require` tags or `globalResources()`) have their styles globally injected *even if they are not used anywhere*.

That in itself doesn't really make sense to me. Css for an element should only be added when that element is added. If adding an element is not what triggers adding the css, then removing an element cannot / should not trigger removing the css either.

Webpack again....

So apart from this particular PR, I feel that this apparently eager loading of all custom element css is an issue in itself that needs to be solved. @EisenbergEffect  do you agree with that or am I seeing this the wrong way?

fkleuver:
To make this work, style injection needs to be changed at a fairly fundamental level. I'm closing this PR because I wrote this code based on a wrong understanding of how style injection currently works, and it won't work like this.  I might have another crack at it at a later time.

jbockle:
I wrote a plugin that addresses this for webpack projects: [aurelia-useable-style-loader](https://github.com/jbockle/aurelia-useable-style-loader)

PR #345, opened 2018-03-31
WIP fix(compose): await prev lifecycle composition fix #335

### LINKED ISSUE #335: `<compose>` inserts contents twice

In certain circumstances, `<compose>` will result in contents of a template being added to the DOM twice. This is because `Compose.bind` and `Compose.unbind` do not wait for previously queued updates to complete before requeueing more updates. Thus, if `bind`, `unbind`, and `bind` are called in quick succession, `CompositionEngine.compose` is called a first time and quickly called a second time, before the first composition has completed. Each call to `compose` queues a call to `ViewSlot.add` before either is complete.

Thus there is a race condition. The second call to `compose` queues a call to `ViewSlot.add` before the first call to `compose`/`ViewSlot.add` has completed. Thus, instead of properly swapping the contents of the view, the templated contents are added twice.

I had [this](https://gist.run/?id=c358985f6c3713e33d566f531f8e9df0) reproducing the bug unreliably but it seems not to be happening now.

**Fix**
To fix this, [`Compose.bind`](https://github.com/aurelia/templating-resources/blob/cd8df6ddf0beab614377ed82001294dc5e24c2d3/src/compose.js#L86) and [`Compose.unbind`](https://github.com/aurelia/templating-resources/blob/cd8df6ddf0beab614377ed82001294dc5e24c2d3/src/compose.js#L98) should be wrapped with `(this.pendingTask || Promise.resolve()).then(() => { ... })` or equivalent.

**Illustration**
The following is not an example. To reproduce this bug, I think the view would need to be far more complex and take far longer to render/build/add. I tried to reproduce the bug with gist.run, but it didn't work.

app.html:
```
<template>
  <compose view="./view.html"></compose>
</template>
```

view.html:
```
<template>
  ... some content ...
</template>
```

result:
```
<main aurelia-start="app">
  ... some content ...
  ... some content ...
</main>
```

* **Library Version:**
1.5.4

* **Operating System:**
Windows 10

* **Node Version:**
7.7.3

* **NPM Version:**
4.1.2

JSPM 0.17.0-beta.47

* **Browser:**
all

* **Language:**
all

**Current behavior:**
When using `<compose>`, the contents of the view gets added to the DOM twice.

**Expected behavior:**
When using `<compose>`, the contents of the view gets added to the DOM once.

StrahilKazlachev:
Await ongoing composition from previous lifecycle when component is rebound - schedule the changes from the new *bind* instead of initiating a new, parallel, composition.


StrahilKazlachev:
@firelizzard18 can you try this? I committed the *dist* also, so you can `npm install` it from here for the test.

bigopon:
Is the new plan let compose process changes continue even if it was unbound ?

firelizzard18:
@StrahilKazlachev I don't have any code at the moment that is reproducing this issue

firelizzard18:
@Alexander-Taran I did not say that I do not have time to test it. I am saying that the app, which is currently using `<compose>`, is no longer exhibiting the issues that prompted me to create the issue. If I knew how to reliably reproduce the issue, I would have done so in the gist.run.

Alexander-Taran:
@firelizzard18 a-ha.. I see.. Ok then (-:

StrahilKazlachev:
OK, so the change I made here is more of a workaround, it will avoid composing in parallel, but does not address the root of the issue - not being able to *cancel* a composition.

EisenbergEffect:
I'll get this merged shortly.

PR #348, opened 2018-04-20
fix(compose): rethrow instead of logging fix #346

### LINKED ISSUE #346: Rethrow errors caught if compose fails
We had some files not deploy correctly and the first thing we knew about it was a report from a tester whereas I would have expected something almost immediately from our error reporting provider.

Reproducing it myself I can see the following in the console which is written to the logger at [this line](https://github.com/aurelia/templating-resources/blob/fdb881733ce41d0ea07b3794d71a03fcfb88db44/src/compose.js#L194).
`ERROR [templating-resources] Error: (SystemJS) XHR error loading blah/blah/blah.js`

Is there any chance this could be rethrown somewhere so anything hooked into handle errors can do its notification business? Logging to console isn't much use to anybody outside of a development environment. 

There are a few other places this happens like in [templating-router](https://github.com/aurelia/templating-router/blob/8d9c7c1bec19aa6f59539012f20e0a9bd2bf1a67/src/route-href.js#L54).

I'll happily do a PR if somebody can suggest the best way to fix this.

StrahilKazlachev:
Rethrow composition errors instead of logging them as errors.
Will result in unhandled Promise rejection.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -8,8 +8,6 @@ import {
 } from 'aurelia-templating';
 import {DOM} from 'aurelia-pal';
 
-const logger = LogManager.getLogger('templating-resources');
-
 /**
 * Used to compose a new view / view-model template or bind to an existing instance.
 */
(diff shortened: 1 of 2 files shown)


timfish:
@StrahilKazlachev this is perfect

aegenet:
Hi,

I understand the need, but for me it's a **breaking change**.

We have an application with a lot of customization (via `compose`). 
`rethrow` the error like this (unhandled promise rejection) seems to **break** our application in Firefox.

My temporary quick fix is to override the `compose` custom element to restore the old behavior (with a small improvement *for our use* : display the error formatted in HTML.

In my opinion, this change should have been activated with a new attribute, as this depends on the use case of the `compose`.

aegenet:
Don't get me wrong, what I'm saying is that it's a breaking change.

Not that it's illogical.

Before: a handled error (ok swallowed, but logged).
After: an error rejected. If the application did not have "unhandled Promise rejection" handling : It will no longer work (in Firefox).

PR #349, opened 2018-05-19
fix(Repeat): ignore mutation while processing instance changed

bigopon:
This PR is for https://github.com/aurelia/framework/issues/830

I tried to reproduce the broken behavior in test but failed.

@EisenbergEffect @jdanyow

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -135,7 +135,11 @@ export class Repeat extends AbstractRepeater {
     if (!this.isOneTime && !this._observeInnerCollection()) {
       this._observeCollection();
     }
+    this.ignoreMutation = true;
     this.strategy.instanceChanged(this, items);
+    this.observerLocator.taskQueue.queueMicroTask(() => {
+      this.ignoreMutation = false;
+    });
   }
 
   _getInnerCollection() {
@@ -151,6 +155,9 @@ export class Repeat extends AbstractRepeater {
   */
   handleCollectionMutated(collection, changes) {
     if (!this.collectionObserver) {
+      return;
+    }
+    if (this.ignoreMutation) {
       return;
     }
     this.strategy.instanceMutated(this, collection, changes);
--- a/test/repeat-integration.spec.js
+++ b/test/repeat-integration.spec.js
@@ -1031,6 +1031,29 @@ describe('Repeat array (pure)', describeArrayTests.bind(this, true));
 
 describe('Repeat array (not pure)', describeArrayTests.bind(this, false));
 
+describe('instancesChanges and instancesMutated together', () => {
+  it('handles  together correctly', () => {
+    let template = `<template><section if.bind='show'><div repeat.for="item of items">\${item}</div></section></template>`;
+    let viewModel = {
+      show: false,
+      // items: [{ a: 'a' }, { a: 'b' }]
+      items: null,
+    };
+    let controller = createController(template, viewModel, false);
+    let taskQueue = container.get(TaskQueue);
+    taskQueue.queueMicroTask(() => {
+      viewModel.show = true;
+      viewModel.items = ['a', 'b', 'c', 'd'];
+      viewModel.items.splice(1, 1, { a: 'bb' });
+      viewModel.items.push({ a: 'aa' });
+    });
+    taskQueue.flushMicroTaskQueue();
+    taskQueue.queueMicroTask(() => {
+      console.log(host.children.length, select(controller, 'div').length);
+    });
+  });
+});
+
 describe('Repeat map [k, v]', () => {
   let viewModel, controller;
   let obj = {};


EisenbergEffect:
@jdanyow @martingust This is a potentially critical fix to the `repeat` functionality in Aurelia. @bigopon has take a pass at the fix but we need another pair of eyes on this, preferably someone who has worked on this code extensively before. Would you one/both of you two be able to look at this PR? If you have any idea on how to get the test to correctly produce the issue, that would be great. Thanks!

bigopon:
Regardless of real source of this issue, I think it's good to have this patched. A simple demo (https://gist.run/?id=d6eef539a65f5ac55bcbeb656eb0f779) to show the differences between Repeat on raw array (without `ignoreMutation` https://github.com/aurelia/templating-resources/blob/master/src/repeat.js#L152-L157) and array after a value converter (thus have `ignoreMutation` (https://github.com/aurelia/templating-resources/blob/master/src/repeat.js#L162-L170)

@EisenbergEffect @jdanyow @fkleuver

fkleuver:
I agree. It's a workaround, but that's better than a flat-out bug. It's probably a good idea to create a new issue for this (or keep the existing one open) so as to make it explicit that this still needs a better fix.

I'm currently working on a different approach altogether but it's a fairly large undertaking. It will take some more time (and I'm not even sure if it will end up working)

bulldetektor:
Any estimates on when we can expect a release with this fix?

EisenbergEffect:
Very soon. We are in the middle of a big release push across many of our libraries. This will come out in the new templating release either this weekend or next week. It will be an RC release at first.

huochunpeng:
@bigopon This PR is the cause of issue(2) in
https://discourse.aurelia.io/t/using-require-breaks-router/1411/14?u=huochunpeng

huochunpeng:
I think there is an opportunity of architecture improvement in aurelia-binding.

Currently all observations (property/collection/dirty/expression) queue micro tasks that directly calls subscribers (which renders updates to DOM).

This is not only inefficient (`this.foo=1; this.foo=2; this.foo=3` will render `<template>${foo}</template>` three 3 times), but also prone to code that mutate model in micro tasks themself (which is the issue that this PR tried to solve).

I would propose that all observations queue micro tasks to register changes, not call subscribers directly.

For instance for property-observation.js
```js
  call() {
    // ...
    // this.callSubscribers(newValue, oldValue);
    this.aureliaRenderOrSomething.registerChange(this, newValue, oldValue);
  }
```

The aureliaRenderOrSomething should merge multiple changes from same observable. For example, if `foo` init value is 0; `this.foo=1; this.foo=2; this.foo=3` will only register one change, not three `(newValue = 3, oldValue =0)`. Beware, this optimisation also merges mutations done by multiple micro tasks.

Merge for CollectionObservation is bit more troublesome but can be done.

When aureliaRenderOrSomething detects first registerd change, it queues a macro task to call subscribers (flush all changes to DOM, **without fear of model mutation during rendering**).

This not only improves performance, but also simplifies code (because rendering is processed with frozen model). Let me know what you think.

fkleuver:
This ^^^^^^
I was planning on implementing exactly that in vNext (thought about calling it `ChangeQueue`).
As for the merging that you're talking about, I already built something like that for `ArrayObserver` in vNext and it's working quite well: https://github.com/fkleuver/experiment/blob/binding-attempt-2/src/runtime/binding/observation/array-observer.ts
It's basically an incremental change tracker.
@huochunpeng  If you want, I'd be more than happy to work together with you on this. There's a lot of ground to cover because it requires a fair amount of rewriting ;)

bigopon:
> The aureliaRenderOrSomething should merge multiple changes from same observable. For example, if foo init value is 0; this.foo=1; this.foo=2; this.foo=3 will only register one change, not three (newValue = 3, oldValue =0). Beware, this optimisation also merges mutations done by multiple micro tasks.

I think this is the current behavior of Aurelia.

fkleuver:
Correct. Observers set their status to `this.queued = true` when they have a change and they don't re-queue on consecutive changes if they're already queued.
Still a separate `ChangeQueue` seems like a good idea. I don't think DOM updates should be in the same task queue as everything else.

huochunpeng:
@fkleuver I am quite happy you do all the hard work :-) I got some idea but it does not necessarily mean I have enough motivation to drive through it.

@bigopon you sure? I will run some test tomorrow.

huochunpeng:
Right. So my understanding is falsy. Thx for clarifying.

Now I can recall where my memory went wrong.

When I worked on optimisation on computed property, I noticed if you use the same property 3 times in view template, the getter will be called 3+1=4 times for one rendering. I was expecting an optimised single getter call for rendering.

fkleuver:
> I am quite happy you do all the hard work :-) I got some idea but it does not necessarily mean I have enough motivation to drive through it.

Would love to hear your ideas :)

Issue #356, opened 2018-07-13
oneTime binding inside if inside repeat does not work as expected


rluba:
**I'm submitting a bug report**

* **Library Version:**
1.7.1

**Please tell us about your environment:**
* **Operating System:**
OSX 10.x

* **Node Version:**
8.9.1

* **NPM Version:**
61.0

* **JSPM OR Webpack AND Version**
 webpack@4.8.3

* **Browser:**
all

* **Language:**
all

**Current behavior:**
Similar to aurelia/framework#301: `repeat.for` and `oneTime` bindings don’t work properly together. 

The specific problem in aurelia/framework#301 is fixed, but as soon as you insert a `if.bind` between the `repeat.for` and the One-Time-Binding, the problem reappears:

welcome.html
```Html
<template>
    <button click.trigger="clickme()">Click me!</button>
    <div repeat.for="item of items">
	    <div if.bind="item">
	    	${item.prop1 & oneTime}
	    </div>
	</div>
</template>
```

welcome.js
```
export class Welcome {
    items = [{prop1:1},{prop1:2},{prop1:3}];

    clickme() {
        this.items = [{prop1:4},{prop1:5}];
    }
}
```

After clicking the button, the content reads
```
1
2
```
i.e. the number of items is updated, but their content is not.

**Expected/desired behavior:**
After clicking the button, the content should be 
```
4
5
```

rluba:
After having a quick look at how repeat works, it is clear why this fails.

repeat simply tries to update all one-time bindings it knows about, but it fails to find one-time bindings within any child views, eg., when using custom elements or `@templateController` attributes like `if`.

The only reliable solution I see is disabling `repeat`s [view-reuse](https://github.com/aurelia/templating-resources/blob/28ceddc40fcb98adc2801e4c772e3df6586299dc/src/array-repeat-strategy.js#L124-L157). It should only reuse views if the view’s bound element actually remains the same. (Similar to how `ngRepeat` in AngularJs works. There you can optionally add a `track by` expression to aid with reusing views, but the default is to throw away any views whose elements have disappeared.)

rluba:
After trying for a few hours, it seems to me like there’s no way to

- either reliably find all one-time bindings that might need updating from [within `repeat.updateBinding`](https://github.com/aurelia/templating-resources/blob/28ceddc40fcb98adc2801e4c772e3df6586299dc/src/repeat.js#L272-L286),
[...]

PR #357, opened 2018-07-13
Fix/repeat one time

rluba:
This is a crude attempt of fixing #356.

The fix for the `if` case was relatively straight-forward. But to disable the fast-pass for the "subview without lifecycle methods" case, I had to rely on `instruction.initiatedByBehavior`, because I couldn't find a better indicator.

I just started working with Aurelia a few weeks ago, so I’m sure there’s a more elegant way that I couldn’t discover. Let me know what you think!

### DIFF
--- a/src/analyze-view-factory.js
+++ b/src/analyze-view-factory.js
@@ -9,7 +9,8 @@ function behaviorRequiresLifecycle(instruction) {
   let name = t.elementName !== null ? t.elementName : t.attributeName;
   return lifecycleOptionalBehaviors.indexOf(name) === -1 && (t.handlesAttached || t.handlesBind || t.handlesCreated || t.handlesDetached || t.handlesUnbind)
     || t.viewFactory && viewsRequireLifecycle(t.viewFactory)
-    || instruction.viewFactory && viewsRequireLifecycle(instruction.viewFactory);
+    || instruction.viewFactory && viewsRequireLifecycle(instruction.viewFactory)
+    || instruction.initiatedByBehavior;
 }
 
 function targetRequiresLifecycle(instruction) {
--- a/src/if-core.js
+++ b/src/if-core.js
@@ -1,3 +1,5 @@
+import { updateBindings } from './repeat-utilities';
+
 /**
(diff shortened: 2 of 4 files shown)


bigopon:
@rluba @EisenbergEffect

I think all the changes are good except using `initiatedByBehavior`. By default checking that will also make it false positive for any custom element / custom attribute that doesn't actually require lifecycle. I think we have to introduce new property on `If` metadata via:

```js
const meta = metada.get(metdata.resource, If);
meta.behaviorRequiresLifecycle = true;
```

and check:

```js
function behaviorRequiresLifecycle(instruction) {
  let t = instruction.type;
  let name = t.elementName !== null ? t.elementName : t.attributeName;
  return lifecycleOptionalBehaviors.indexOf(name) === -1 && (t.handlesAttached || t.handlesBind || t.handlesCreated || t.handlesDetached || t.handlesUnbind)
    || t.viewFactory && viewsRequireLifecycle(t.viewFactory)
    || instruction.viewFactory && viewsRequireLifecycle(instruction.viewFactory)
    || instruction.type.behaviorRequiresLifecycle;
}
```

Note: naming is for discussion.

src/analyze-view-factory.js
```diff
@@ -9,7 +9,8 @@ function behaviorRequiresLifecycle(instruction) {
   let name = t.elementName !== null ? t.elementName : t.attributeName;
   return lifecycleOptionalBehaviors.indexOf(name) === -1 && (t.handlesAttached || t.handlesBind || t.handlesCreated || t.handlesDetached || t.handlesUnbind)
     || t.viewFactory && viewsRequireLifecycle(t.viewFactory)
-    || instruction.viewFactory && viewsRequireLifecycle(instruction.viewFactory);
+    || instruction.viewFactory && viewsRequireLifecycle(instruction.viewFactory)
+    || instruction.initiatedByBehavior;
```
bigopon: We can not do this, as it includes all custom element, and custom attribute, thus could possibly make it a breaking change.

PR #364, opened 2019-01-24
Ability to turn of caching of view instance when using if

### LINKED ISSUE #320: Allow the possibility to specify the cache type on the `if` attribute

* **Library Version:**
1.5.1

**Current behavior:**
This is a follow up for aurelia/framework#602.

A custom element doesn't get destroyed/re-constructed when used with `if` because the model it's being cached.

After a lot of discussion, we agreed that the feature would be for the `if` to support the following syntaxis:

```
if="value.bind: someValue; cache: none"
```
That way you can specify if the model should be cached or not.

* **Notes:**
  - This can't be specified on the custom element definition because of the way the cache works: it doesn't check element by element, so in a big tree, it should now check one by one, which would be bad for performance.
  - It can't be done at an application level because if you use third party libraries, it would affect them, changing their implementations of the `if` and possibly causing them to malfunction.

fracz:
fix #320 

I believe the syntax of this `if` is:

```html
if="condition.bind: someValue; cache: false"
```

Don't know if the CI fail is due to my changes. All seems good locally.

### DIFF
--- a/src/if-core.js
+++ b/src/if-core.js
@@ -12,6 +12,7 @@ export class IfCore {
     // state anymore, so we use `showing` for that.
     // Eventually, `showing` and `value` should be consistent.
     this.showing = false;
+    this.cache = true;
   }
 
   bind(bindingContext, overrideContext) {
@@ -67,7 +68,7 @@ export class IfCore {
     return this.viewSlot.add(this.view); // Promise or void
   }
 
-  _hide(cacheView = true) {
+  _hide() {
     if (!this.showing) {
       return;
     }
@@ -77,17 +78,15 @@ export class IfCore {
 
     if (removed instanceof Promise) {
       return removed.then(() => {
-        this._unbindView(cacheView);
+        this._unbindView();
       });
     } else {
-      this._unbindView(cacheView);
+      this._unbindView();
     }
   }
 
-  _unbindView(cache = true) {
-    if (cache === 'false') {
-      cache = false;
-    }
+  _unbindView() {
+    const cache = this.cache === 'false' ? false : !!this.cache;
     this.view.unbind();
     if (!cache) {
       this.view = null;
--- a/src/if-core.js
+++ b/src/if-core.js
@@ -67,7 +67,7 @@ export class IfCore {
     return this.viewSlot.add(this.view); // Promise or void
   }
 
-  _hide() {
+  _hide(cacheView = true) {
     if (!this.showing) {
       return;
     }
@@ -76,9 +76,17 @@ export class IfCore {
     let removed = this.viewSlot.remove(this.view); // Promise or View
 
     if (removed instanceof Promise) {
-      return removed.then(() => this.view.unbind());
+      return removed.then(() => {
+        this.view.unbind();
+        if (!cacheView) {
+          this.view = null;
+        }
+      });
+    } else {
+      this.view.unbind();
+      if (!cacheView) {
+        this.view = null;
+      }
     }
-
-    this.view.unbind();
   }
 }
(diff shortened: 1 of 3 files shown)


bigopon:
`cache: false` will give a literal string `false` to `cache` property.
`cache.bind: false` will give false the boolean to `cache` property

I think it's a potential issue/confusion. Though I'm not sure if it should automat(g)ically convert `'false'` to `false`

@EisenbergEffect @fkleuver @jdanyow @jods4

EisenbergEffect:
It's less clear in vCurrent but I think in this case we should probably auto-convert it if we get a string. Frankly, the basic `if` value probably suffers the same issue.

fracz:
Updated to support `cache: false` as string.

src/if.js
```diff
@@ -12,6 +12,7 @@ import {IfCore} from './if-core';
 export class If extends IfCore {
   @bindable({ primaryProperty: true }) condition: any;
   @bindable swapOrder: "before"|"with"|"after";
+  @bindable cache: boolean = true;
```
bigopon: I think if the value can potentially be string, we should be a more correct about the type here `boolean | string`

src/if-core.js
```diff
@@ -67,7 +67,7 @@ export class IfCore {
     return this.viewSlot.add(this.view); // Promise or void
   }
 
-  _hide() {
+  _hide(cacheView = true) {
```
bigopon: I think it's easier to define a `cache` property in base class for defining caching behavior instead of passing in as an argument. As by the time an animation has finished, the `cache` value may have been changed. @EisenbergEffect @jods4 @fkleuver

EisenbergEffect: Don't we have an issue with bindable inheritance?

bigopon: No if bindable is declared on derived class. It's a requirement, not an issue really.

About the above, it's just property value/definition on the base class, so it can be used with `this.cache` to always use the latest value to determine if it should throw away the view

EisenbergEffect: Ah. Gotcha. That makes sense. I think that would be better and you'd always be sure to have the latest value.

fracz: Not sure if this is what you were asking fore - but here you go!

bigopon: @fracz  Nice work. That's it. Now... It would be awesome if there could be follow up PR in doc about this new feature, and 2 or some more tests for the positive cases, where `view !== null` 🚀

EisenbergEffect: Good points. One of the challenges we've had is making nice improvements like this but not getting them properly documented. If we can add the tests here and then a small bit in the documentation here: https://github.com/aurelia/documentation/blob/master/current/en-us/3.%20fundamentals/5.%20cheat-sheet.md I think we'd be ready to rock and roll. I won't block merging this for the docs. If we can get the tests in that would be good to get this merged.

fracz:
PR with docs has been added https://github.com/aurelia/documentation/pull/380 Now you can be sure I'm desperate.

As for the "positive" tests - they are already there. Although they not explicitly indicate, [they are asserting that `view` is there after being unbound](https://github.com/aurelia/templating-resources/blob/6e7520c157186a1f74cc95c71700bcb08fdfdaea/test/if.spec.js#L154). Not sure if you want me to repeat the same test, but with explicit name telling that we test the default caching behavior.

bigopon:
@fracz ahah, awesome work. this looks great to me 😄 

About the tests, I think it's fine 👍 
@EisenbergEffect

EisenbergEffect:
I'll merge the docs next. This release should go out today or tomorrow. Docs will probably push out next week.

homer0:
@fracz thank you!!! I've been waiting a long time for this too :D

PR #363, opened 2019-01-19
fix(ci): update deps, jspm deps, test deps, ci

bigopon:
@EisenbergEffect @fkleuver

PR #368, opened 2019-03-06
feat(html-resource-plugin, dynamic-element): Add shadow dom support

### LINKED ISSUE #367: [Feature request] Support useShadowDOM for html only components

* **Library Version:**
1.8.0

**Current behavior:**
It is not possible to use shadow dom on html only custom components.

It would be nice to support shadow dom on html only components.
Proposal:
`<template use-shadow-dom>`

Less files 🎉 easier to maintain, faster to write.

michaelw85:
Introduce attribute use-shadow-dom for html only components.

Closes #367


EisenbergEffect:
@bigopon Can you review this? @fkleuver I think we'd want to add this to vNext template compiler as well, so that it extracts this setting from HTML if present on a template element.

fkleuver:
@EisenbergEffect Could you create an issue / TODO for that? There was something else a few days ago as well. I can't really keep track of things like this otherwise 😛

bigopon:
The change related to new feature lgtm. There are no tests, though we can still merge and I'll write it, or ask @michaelw85 to help a bit more with them :) @EisenbergEffect @fkleuver

EisenbergEffect:
Should we enable specifying open or closed mode?

`use-shadow-dom="open"` and `use-shadow-dom="closed"`

We can also make `use-shadow-dom` a shortcut for `use-shadow-dom="open"`

bigopon:
@EisenbergEffect right, I forgot that 😄 , and agree

fkleuver:
I'm a little concerned about the (dated) CI changes that made their way into this PR. For one, it's yet more deviation, and also some aurelia dependencies were updated. Are we sure this is going to make for a stable release?

bigopon:
@fkleuver for CI, those config should be removed and `cut-release` is the only thing needed?

fkleuver:
@bigopon You can see [here](https://github.com/aurelia/circleci/blob/master/src/v1.yml#L8) what assumptions it makes. There needs to be a build, test and release script. The defaults are `npm run build`, `npm run test`, `npm run cut-release`. They can easily be overridden as long as there are 3 such scripts. It also assumed package-lock to be up to date (as it uses `npm ci`) and for there for be some jspm config.

michaelw85:
> The change related to new feature lgtm. There are no tests, though we can still merge and I'll write it, or ask @michaelw85 to help a bit more with them :) @EisenbergEffect @fkleuver

@bigopon I wanted to write a test and noticed none existed yet. I figured since I primarily add existing (proven?) functionality, the effort to figure out how to and get this tested vs risk, was not worth it. If you can easily add a test that would be great although I think I should do it since I created the PR.

> Should we enable specifying open or closed mode?
> 
> `use-shadow-dom="open"` and `use-shadow-dom="closed"`
> 
> We can also make `use-shadow-dom` a shortcut for `use-shadow-dom="open"`

@EisenbergEffect I am unaware of such options, is there any documentation explaining what these options do and how to apply them?

> I'm a little concerned about the (dated) CI changes that made their way into this PR. For one, it's yet more deviation, and also some aurelia dependencies were updated. Are we sure this is going to make for a stable release?

@fkleuver I wanted to provide a PR with a passing CI build but I was wrong to merge another PR into mine. Do you agree if I revert the merge and this discussion continues in the original PR #363

bigopon:
> @EisenbergEffect I am unaware of such options, is there any documentation explaining what these options do and how to apply them?

you can do the following:
```html
<template use-shadow-dom="closed">
```
translated to:
```js
@useShadowDOM({ mode: 'closed' })
class MyEl {
  ...
}
```

> @bigopon I wanted to write a test and noticed none existed yet. I figured since I primarily add existing (proven?) functionality, the effort to figure out how to and get this tested vs risk, was not worth it. If you can easily add a test that would be great although I think I should do it since I created the PR.

I can add few basic tests related to this, and then if you are still keen, we can collaborate on this afterwards.

michaelw85:
@bigopon Thanks I will update the PR with the open/closed options. 
I was thinking to adding the following scenarios:
1) No explicit setting `use-shadow-dom` will use the decoration without options this will make the decorator responsible for setting the default. If for any reason the default would change in the future the attribute and decorator will act identical.

2) explicit (open/closed) passing the settings object to the decorator.

EisenbergEffect:
Sounds great!

fkleuver:
@michaelw85 No, I think it was a smart move to merge it so that you could get CI to pass. I'm just pointing out that, when your PR itself is done, we may either want to split off the CI/workflow stuff into a different PR and merge that first so that the scope stays manageable. There are a few things in that PR which I'm not very comfortable with, mainly the major version bump on `aurelia-bootstrapper`. We have gotten quite some trouble from major version bumps before.

michaelw85:
Ok, if this PR will not merge any CI changes I still think #363 should be finished and your comments should be resolved in that PR. I can rebase on master before this gets reviewed and merged keeping things clean an separated.

bigopon:
Re: major version bump of `aurelia-bootstrapper`: it's completely safe as it's dev dependencies in jspm packages. It won't affect end users of this package in anyway. About the changes related to new feature, it lgtm. I think it's safe to go.
@fkleuver @EisenbergEffect

michaelw85:
Updated code to allow explicitly passing shadow mode mode open/closed as requested by @EisenbergEffect 
Updated passing settings from separated params to a object using object destructuring as discussed with @bigopon 

----
Trying to make these changes I had a couple difficulties with my setup.
1) vscode is not happy about adding typings to js files. How should I configure this correctly or use this?
2) I have an eslint extension installed but it does not seem to autofix with this repo. Should I do any specific configuration? (Previous question might be the reason that it is not formatting)
3) To create these changes I tried to npm install my local dev branch of templating resources in a cleanly installed aurelia project. This completely failed. I had the same issues when testing aurelia-store recently. 
I ended up just copy/pasting dist files to be able to manually test. Is there any procedure how to create a local aurelia project using local deps for development?
4) After running gulp build everything in dist is shown as changed... I just ignored it for now but should  the dist folder be added to the gitignore?

src/dynamic-element.js
```diff
@@ -1,15 +1,40 @@
...
+  case 'closed':
+    useShadowDOM({ mode: 'closed' })(DynamicElement);
+    break;
+
+  case '':
+    useShadowDOM(DynamicElement);
+    break;
+
+  default:
+    // Do not use shadow dom
```
bigopon: I think there should be a warning here
```js
import { getLogger } from 'aurelia-logging';

    // Do not use shadow dom
    getLogger('aurelia-html-only-element')
      .warn(`Expected 'use-shadow-dom' value to be "close", "open" or "", received ${useShadowDOMmode}`)
```

src/html-resource-plugin.js
```diff
@@ -6,23 +6,24 @@ export function getElementName(address) {
...
   viewEngine.addResourcePlugin('.html', {
-    'fetch': function(address) {
-      return loader.loadTemplate(address).then(registryEntry => {
-        let bindable = registryEntry.template.getAttribute('bindable');
-        let elementName = getElementName(address);
+    'fetch': function(viewUrl) {
+      return loader.loadTemplate(viewUrl).then(registryEntry => {
+        let bindableNames = registryEntry.template.getAttribute('bindable');
+        const useShadowDOMmode: null | '' | 'open' | 'closed' = registryEntry.template.getAttribute('use-shadow-dom');
+        const name = getElementName(viewUrl);
```
bigopon: Lol, you renamed `bindable` to `bindableNames` while renaming `elementName` to `name`? 😆 

Let's have `elementName`, I think it's general theme of naming convention in Aurelia.

bigopon:
`dist` files are kind of weird case, I'm not sure what to do with it. It's kind of annoying for contributors sometimes. Maybe @fkleuver can config it right? 

About (3), I'm not sure why, but I can have a look later.

src/html-resource-plugin.js
michaelw85: I matched the values with params in dynamic element. I would like to keep the names identical, should I update both or only here?

src/dynamic-element.js
michaelw85: ok I could do that. I will also have to include a undefined case that does nothing to be able to default to a warning.

michaelw85:
> `dist` files are kind of weird case, I'm not sure what to do with it. It's kind of annoying for contributors sometimes. Maybe @fkleuver can config it right?
> 
> About (3), I'm not sure why, but I can have a look later.

I would assume a local npm install of package would work out of the box. I don't understand what's different compared to a package pulled down via npm.

src/html-resource-plugin.js
bigopon: I missed that, yeah lets keep it this way. Just small nit anyway 👍

michaelw85: ok I won't change the naming and will close this conversation.

src/dynamic-element.js
michaelw85: Added warning + null case when attribute is not set

EisenbergEffect:
Looks ready to merge. Any objections?

fkleuver:
Unless there is time pressure to get this released, it'd be nicer to have this PR split up into one that fixes the workflow, and one that addresses this issue. That workflow stuff was WIP and would not bring things in a consistent state right now.

michaelw85:
@fkleuver Reverted merged CI build changes.

PR #369, opened 2019-03-18
 fix(ci): fix circle ci config

michaelw85:
Config based on aurelia CLI setup + old setup.


michaelw85:
Some additional info on files changed:
- config.js 
> "npm:aurelia-binding@1.6.0" was automatically removed after performing jspm install.
- package.json jspm 
> peer deps also automatically removed after jspm install.
- repeat-integration.spec.js
>Removal of test since it failed locally and on build. After looking at the test I noticed it does not have any assertions so I decided to remove it in favor of getting CI builds up and running again.

michaelw85:
@fkleuver Here's a PR to get the CI working again. I tried to keep the amount of changes/impact to a minimum.

@EisenbergEffect It might be good to get this merged prior to #368

test/repeat-integration.spec.js
```diff
@@ -1031,29 +1031,6 @@ describe('Repeat array (pure)', describeArrayTests.bind(this, true));
...
-      viewModel.items.splice(1, 1, { a: 'bb' });
-      viewModel.items.push({ a: 'aa' });
-    });
-    taskQueue.flushMicroTaskQueue();
-    taskQueue.queueMicroTask(() => {
-      console.log(host.children.length, select(controller, 'div').length);
-    });
-  });
-});
-
```
fkleuver: Why is this test removed? If it fails for known reasons, better to skip it with `xdescribe` and add a reason

package.json
```diff
@@ -22,23 +22,11 @@
...
-      "aurelia-binding": "^2.0.0",
-      "aurelia-dependency-injection": "^1.0.0",
-      "aurelia-loader": "^1.0.0",
-      "aurelia-logging": "^1.0.0",
-      "aurelia-metadata": "^1.0.0",
-      "aurelia-pal": "^1.3.0",
-      "aurelia-path": "^1.0.0",
-      "aurelia-task-queue": "^1.0.0",
-      "aurelia-templating": "^1.8.1"
-    },
```
fkleuver: What does removing these peer deps solve?

.circleci/config.yml
```diff
@@ -0,0 +1,31 @@
...
+      - run: ./node_modules/.bin/jspm install
+
+      - save_cache:
+          paths:
+            - node_modules
+          key: node10-dependencies-{{ checksum "package.json" }}
+
+      # run tests!
+      - run: node ./node_modules/gulp/bin/gulp.js build
+      - run: node ./node_modules/gulp/bin/gulp.js test
```
fkleuver: Could reuse the [orb](https://github.com/aurelia/circleci/blob/master/src/v1.yml#L8) here if you just add package.json scripts that map to these gulp commands as well as prepare-release, so in package.json:

```json
"scripts": {
  "test": "gulp test",
  "build": "gulp build",
  "cut-release": "gulp prepare-release"
}
```

And then the entire config.yml can be reduced down to this:

```yml
version: 2.1

orbs:
  v1: aurelia/v1@volatile

workflows:
  main:
    jobs:
      - v1/build_test
```

We are trying to normalize all vCurrent repos like this (about 18 of the 40 or so repos are currently migrated to this new workflow). So if you're willing to just make this change to this particular repo then we'll be on 19 :)

If you ran out of time that's np, I'll just do it later then.

michaelw85: Ok that sounds like a great improvement. I will give it a shot. I noticed the other PR for fixing the CI was merged so it might take me a bit more time to get things in order. I will close this PR and start a new one instead.

michaelw85:
Abandoned since #363 has been merged.

Issue #295, opened 2017-07-25
Matcher binding on repeat elements is only working the first time the element is attached


suneelv:
**I'm submitting a bug report**

* **Library Version:**
1.4.0

**Please tell us about your environment:**
* **Operating System:**
Same issue on both OSX and Windows10

* **Node Version:**
6.2.0

* **NPM Version:**
3.8.9

* **JSPM OR Webpack AND Version**
webpack 3.3.0

* **Browser:**
all 

* **Language:**
ESNext

**Current behavior:**

Array repeat strategy is losing the matcher binding once the element is removed from the view. The function bound to `matcher` on repeat is only called when the element is rendered the first time. Once the parent element (element which contains repeat elements) is removed from dom (because of if binding/route change) and attached to the dom again, the matcher is not called from here on.

**Expected/desired behavior:**

https://gist.run/?id=fe857a95d87d2f07bd5a8ee86e1fd15e

When the element is attached for the first time, scope has a matcher
![image](https://user-images.githubusercontent.com/6492948/28590192-8f4d0952-7181-11e7-9cad-9f79f8119c41.png)

When the element is attached again later, scope has no matcher
![null matcher](https://user-images.githubusercontent.com/6492948/28994809-7974da16-79d7-11e7-927c-040b559c0933.png)

* **What is the expected behavior?**
The function bound to matcher should be called everytime

suneelv:
I believe the culprit is 

```
  _captureAndRemoveMatcherBinding() {
    if (this.viewFactory.viewFactory) {
      const instructions = this.viewFactory.viewFactory.instructions;
      const instructionIds = Object.keys(instructions);
      for (let i = 0; i < instructionIds.length; i++) {
        const expressions = instructions[instructionIds[i]].expressions;
        if (expressions) {
          for (let ii = 0; i < expressions.length; i++) {
            if (expressions[ii].targetProperty === 'matcher') {
              const matcherBinding = expressions[ii];
              expressions.splice(ii, 1);
              return matcherBinding;
            }
          }
        }
      }
    }

    return undefined;
  }
```
in https://github.com/aurelia/templating-resources/blob/master/src/repeat.js
This is removing the matcher expression from the viewFactory if it is found. But I guess viewFactory is cached and the next time the element is bound, the `matcher` expression won't be there

EisenbergEffect:
Should be fixed by this: https://github.com/aurelia/templating-resources/pull/372

Seebiscuit:
[...]

PR #372, opened 2019-04-11
Update repeat.js

HubertoKusters:
fix(templating-resources): change indexer in repeat.js(_captureAndRemoveMatcherBinding)

the inner for-loop uses the indexer of the outer for-loop which is incorrect

breaking changes: None


bigopon:
@HubertoKusters Wow that's a big oopsie. Thanks for spotting and fixing this.

cc @EisenbergEffect 

Could be relevant: https://github.com/aurelia/templating-resources/issues/295

Issue #222, opened 2016-05-11
Composing a view can access the parent property


leon-andria:
We have 2 views: main.html and child.html.
Inside main.html, we do a compose on child.html:

```
 <compose view-model="child"></compose>
```

Note that no model is passed.
Now, in main.js, we specify in the activate method:

```
this.manouche = "Hello";
```

Now, in child.html:

```
  <p>${manouche}</p>
```

The result is that in the child.html, we got "Hello".
Does this is an expected behavior or it's a bug ? This means Compose can inherits from its parent ?

alvarezmario:
@leon-andria yep, compose inherits from its parent. In any case, someone from the core team should be more suitable for the final answer

alvarezmario:
in fact, according to this https://github.com/aurelia/templating-resources/blob/master/src/compose.js#L73 is in fact the expected behavior

EisenbergEffect:
Composing a view automatically inherits its binding context from the parent if no view-model is provided, otherwise there would be nothing to bind to. If this is not desired, you can provide a fake view model like this `view-model.bind="{}"`. I haven't tried that myself, but it should work. Let me know if it doesn't.

leon-andria:
@Rob, thanks for your clarification, it's what I expected before to ask if it's a bug or not ?
So, with the following code:
```
<compose view-model.bind="child"
                 model.bind="model">
```

In my child, i can display "Hello".
This means that Compose doesn't isolate anymore ? Is it a design choice ?

EisenbergEffect:
Yes, the compose element enables flowing through the binding context and override context. This is different from standard components which are fully encapsulated. This means that with compose, if the binding engine doesn't find a match on the composed view model, it can walk the binding context tree. (Traditional components can't because the trees aren't connected.)

We wanted to provide this as a possibility for certain scenarios where it made sense. However, it's not advisable to take advantage of this except in rare cases, since it creates unnecessary coupling from child to parent.

If you would like an option to explicitly prevent this, we could extend the compose element with a property "inherit-context" which could be used to turn this off. It would be very simple to add. In the link referenced above we would simply not pass the context of the compose element along to the composition engine. If we don't do that, it cannot connect them.

Let me know your thoughts on this. Also, I'd love to hear from @alvarezmario  and @jdanyow

leon-andria:
@Rob,
[...]

PR #374, opened 2019-04-23
chore(all): convert to TS

### LINKED ISSUE #300: repeat.spec.js tests repeat object from jspm package instead of one from the project source

### LINKED ISSUE #321: cleanup: replace oneWay with toView, add toView + fromView binding behavior
We have some binding behaviors that set the binding mode. Now that we have toView and fromView binding modes we should add the corresponding binding behaviors:

https://github.com/aurelia/templating-resources/blob/0cef07a8cac8e99146d8e1c4b734491bb3dc4724/src/binding-mode-behaviors.js

Also need to standardize the internals on `toView` now that it has landed. We still need to support `oneWay` but the internals should use `toView` where possible.

related to https://github.com/aurelia/binding/issues/629 and https://github.com/aurelia/templating-binding/issues/118

### LINKED ISSUE #332: tests for repeat is not testing the source code in src/ folder.

* **Library Version:**
1.5.4 (this should affect all versions since the beginning)

* **Operating System:**
OSX 10.13

* **Node Version:**
8.5.0

* **NPM Version:**
5.5.1

JSPM 0.16.53

* **Browser:**
NA

* **Language:**
NA

**Current behavior:**
I was investigating the ci test failure on latest v1.5.4 build.
It's a failure test about `unbind()` call for `repeat`. It turns out it is NOT testing the latest source code in src/ folder. It was testing against the aurelia-templating-resource package (very very old version) installed by jspm.

It was all started with this line in test.
```javascript
component = StageComponent.withResources().inView('<div repeat.for="item of items"></div>').boundTo({ items: [] });
```

`withResources()` has no input, it just uses what aurelia loaded, which is all custom attributes coming from the aurelia-templating-resource package installed by jspm.

I tried to use `withResources('src/repeat')`, but aurelia rejected my attempt by saying 
```
Attempted to register an Attribute when one with the same name already exists. Name: repeat.
```
which is reasonable.

I tried to created a `test/repeat-test.js` to extend Repeat (just like `test/if-test.js` did) to hook up local test to `repeatt.for` instead of `repeat.for`, but could not pass the line
```javascript
this.sourceExpression = getItemsSourceExpression(this.instruction, 'repeat.for');
```
in original `src/repeat.js`.
ES6 doesn't allow me to skip super(...) in RepeatTest constructor, which runs the above line I need to 
skip and override with `getItemsSourceExpression(this.instruction, 'repeatt.for');`

Need to find a way for all tests around `repeat` to against local source code, not against package installed by jspm.

bigopon:
Beside conversion, also add ability to configure `Compose` behavior to not inherit binding context by default via entry `configure` function:

```ts
export function configure(config: any, { useLegacyBehavior = true } = {}); {
  Compose.useLegacyBehavior = useLegacyBehavior;
}
```

It's based on @jdanyow 's comment here https://github.com/aurelia/templating-resources/issues/222#issuecomment-294858777

Though I'm not sure when `Compose` is configured to not inherit binding context, it should throw or just silently inherit when composed with view only instruction.

Besides that, adjust build, so now distribution will be a single file with sourcemap enabled. Something like this
image

@EisenbergEffect @jdanyow @jods4 @fkleuver 

@fkleuver What should i adjust circle CI of this to?

@huochunpeng 
resolves #332 
resolves #300
resolves #321


EisenbergEffect:
Wow! @bigopon You are a beast! This is awesome. How should we proceed. I'm not sure if I have a ton of time to review every line. I trust your work though :) Is there anything in particular we should look at, any concerns, details left to polish up?

bigopon:
@EisenbergEffect The plan is like this:

in `templating-router` and `templating-resources`, accept 2nd paramter for `configure` function to toggle legacy behavior for `Compose` and `RouterView`.

in `framework`, add method `enableLegacyBehavior(enabled: boolean)` to `FrameworkConfiguration` to pass that to those two modules.

Write some doc for this.

Details to look at:

There are two way to toggle `Compose` behavior:
- at start up time: like explained above
- at any time: `Compose.useLegacyBehavior = true/false` (default is `true` for compat). I'm not sure about naming.

bigopon:
About the conversion, it's not that much of a big deal I think. Most typings are still kept any for safety reasons. Mostly it's the build/tests/development flow that are cleaner now

EisenbergEffect:
@bigopon Instead of using the generic terms "legacy behavior" can we instead make the property name more explicit in what it's doing? For example `allowParentScopeTraversal = true` or something like that. Not sure exactly what the name would be.

bigopon:
@EisenbergEffect that sounds good to me. Will need to adjust router view accordingly. What about view only composition, do we want to auto traverse parent, or throw?

EisenbergEffect:
For view-only composition, remind me what happens with the binding context. Does it get nothing if it doesn't traverse? That doesn't seem right. Is there a reason we would treat this scenario differently?

bigopon:
It points to the `Compose` element with its `overrideContext` https://github.com/aurelia/templating-resources/blob/6aa889b2f8aa92ca6a9b8acb3a0ba4bdca7372d7/src/compose.ts#L227L228

Which I think it will automatically inherit owning view of `compose` by default. So only need some tests, no need to change anything here

EisenbergEffect:
If it inherits the owning view, that sounds ideal.

bigopon:
@EisenbergEffect here are the two skipped tests
* https://github.com/aurelia/templating-resources/pull/374/files/6aa889b2f8aa92ca6a9b8acb3a0ba4bdca7372d7..5347777153b3ee4b8b90a63fe570d72910827f8f#diff-342077e7f5694040ec714108a9a1234aR134 
* https://github.com/aurelia/templating-resources/pull/374/files/6aa889b2f8aa92ca6a9b8acb3a0ba4bdca7372d7..5347777153b3ee4b8b90a63fe570d72910827f8f#diff-342077e7f5694040ec714108a9a1234aR150

I just realized when adding some integration tests for compose, that `<router-view/>` and `<compose/>` do composition differently. `<router-view/>` controls the behavior instruction, while `<compose/>` does not.

For this, I think I'll have to add something in templating.

EisenbergEffect:
@bigopon I'd rather just delete the code than comment it out usually. Was there a strong argument for doing it this way?

bigopon:
@EisenbergEffect Removed

bigopon:
@EisenbergEffect one of throttle tests is failing, its' probably non-deterministic timing related, we can fix that later.

PR #379, opened 2019-05-23
fix(repeat): properly extract matcher binding

### LINKED ISSUE #378: matcher.bind does not work inside of repeat

* **Library Version:**
1.10.0

* **Operating System:**
Windows 10

* **Node Version:**
10..0

* **NPM Version:**
6.1.0

webpack 4.31.0

* **Browser:**
Chrome

* **Language:**
TypeScript

**Current behavior:**
matcher.bind does not work inside of repeat.
It works with no problem in 1.9.1.

DEMO: https://codesandbox.io/s/objects-uc90f?fontsize=14&module=%2Fsrc%2Fapp.ts
Forked from the demo at https://aurelia.io/docs/binding/radios.
The matcher works inside of repeat.

bigopon:
At the moment, it simply walks through the instruction list to get the right matcher binding, which will break in scenario where folks have bindings with matcher in it, for example: a list of input box:
```html
<label repeat.for="item of items">
  <input matcher.bind="itemMatcher" />
</label>
```
Current code will mis-identify `matcher.bind="itemMatcher"` as repeat matcher.

Fixes #378 

need to investigate #295

---
Explanation:

Every repeat by default looks for a `matcher` binding by walking through view factory instructions and looking for a binding expression targeting a property with name `matcher`. What our current implementation does not do is distinguishing between a `matcher` binding expression on the repeated element, and any element inside it. So for the following template:

```html
<div repeat.for="product of products"
  matcher.bind="product_matcher_by_id">
  <!-- a custom element with inner structure like this -->
  <label>
    <input type="radio"
      matcher.bind="tag_matcher_by_id"
      model.bind="tag"
      checked.bind="product.selectedTag">
      ${tag.name}
  </label>
</div>
```

it works find, but it fails on the following template:

```html
<div repeat.for="product of products">
  <label>
    <input type="radio"
      matcher.bind="tag_matcher_by_id"
      model.bind="tag"
      checked.bind="product.selectedTag">
      ${tag.name}
  </label>
</div>
```
as it gets the `matcher` binding on the `<input/>` element.

### DIFF
--- a/src/array-repeat-strategy.ts
+++ b/src/array-repeat-strategy.ts
@@ -1,5 +1,6 @@
 import {createFullOverrideContext, updateOverrideContexts, updateOverrideContext, indexOf} from './repeat-utilities';
 import {mergeSplice} from 'aurelia-binding';
+import { Repeat } from './repeat';
 
 /**
  * A strategy for repeating a template over an array.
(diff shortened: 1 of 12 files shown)


bigopon:
I just realized the potential breaking change in this fix. Currently we find the first matcher binding and use it as the matcher for repeat. After this fix, we apply some check to make sure we find the right matcher binding, which should be declared on the repeated element, not any element inside it:

```html
<div repeat.for="item of items">
  <input type=radio matcher.bind="itemMatcher">
</div>
```
Current behavior: use `matcher.bind="itemMatcher"` on `<input/>` to compare item for the repeat.
Fixed behavior: there is no matcher.

Maybe we can have a flag on Repeat constructor to toggle the behavior, to avoid a breaking change.

@EisenbergEffect @fkleuver thoughts?

bigopon:
cc @Ischa

bigopon:
cc @Ischa @jods4 @jdanyow

bigopon:
I have added a static property on `Repeat` class named: `useInnerMatcher`, with default value true:
```js
export class Repeat {
  static useInnerMatcher = true;
}
```
So to fix the issue in #378, simply do:

```js
import { Repeat } from 'aurelia-templating-resources'

Repeat.useInnerMatcher = false;
```

Thoughts?

EisenbergEffect:
@bigopon This is marked WIP. Just wanted to make sure you still had stuff you were doing with this... Thanks for thinking of backwards compat :)

Vheissu:
I can confirm this PR solves the issue which I just encountered and needed to have working. Brilliant work.

Vheissu:
@EisenbergEffect Can we get this one merged in please? I've been using it for the last 4 days and it's stable, working well.

EisenbergEffect:
Need @bigopon to remove the WIP designation.

bigopon:
Will have a look again soon

bigopon:
@Vheissu @EisenbergEffect ready for a final review.

PR #382, opened 2019-07-03
feat(compose): bindable activation-strategy

### LINKED ISSUE #381: [Feature request] Make caching of compose configurable

* **Library Version:**
1.11.0

* **Operating System:**
Windows 10

* **Node Version:**
10.15.0

* **NPM Version:**
6.4.1

webpack 4.35.0

* **Browser:**
all 

* **Language:**
TypeScript 3.5.2

**Current behavior:**
The view-model and view used in `compose` is currently cached, that is whenever the model changes, the view-model and view is never recreated and just reactivated with the new model. This makes sense for most of the cases. However, this causes problem for some cases. For example, when validation is used in a custom element. When the model is changed, old validation rules still gets triggered. Some workarounds in this case can be to use an explicit `signal` behavior on `compose` to force recreation of VM, or to reset the validation rules somehow (I tried the later, but without success). For more details on the use case, please refer [this](https://discourse.aurelia.io/t/how-recreate-view-and-vm-for-compose/2685).

Naive expectation is that the VM and View will be recreated, considering all custom elements are transient. However, in this case I would suggest for a `@bindable cache:boolean` flag which can be used in [`processChanges`](https://github.com/aurelia/templating-resources/blob/master/src/compose.ts#L225) to force the recreation. This is similar to the [`cache` property of `if.bind`](https://github.com/aurelia/templating-resources/pull/364). 

Any workaround will take considerable time on my part due to non-trivial refactoring, whereas the solution with the configurable `cache` looks very much easily doable.

If you are interested for a PR, I would like to give it a try.

Sayan751:
This PR introduces bindable activation-strategy for compose. The default strategy is 'invoke-lifecycle', which is also the current strategy. The strategy 'replace' forces re-creation of bound view and view-model on model change.

Fixes #381

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -3,7 +3,22 @@ import { DOM } from 'aurelia-pal';
 import { TaskQueue } from 'aurelia-task-queue';
 import { bindable, CompositionContext, CompositionEngine, customElement, noView, View, ViewResources, ViewSlot } from 'aurelia-templating';
 
-
+/**
+ * Available activation strategies for the view and view-model bound to compose
+ *
+ * @export
+ * @enum {string}
+ */
+export enum ActivationStrategy {
+  /**
+   * Default activation strategy; the 'activate' lifecycle hook will be invoked when the model changes.
+   */
+  InvokeLifecycle = 'invoke-lifecycle',
+  /**
+   * The view/view-model will be recreated, when the "model" changes.
+   */
+  Replace = 'replace'
+}
 
 /**
  * Used to compose a new view / view-model template or bind to an existing instance.
@@ -38,6 +53,15 @@ export class Compose {
    * @type {Class}
    */
   @bindable viewModel: any;
+
+  /**
+   * Strategy to activate the view-model. Default is "invoke-lifecycle".
+   * Bind "replace" to recreate the view/view-model when the model changes.
+   *
+   * @property activationStrategy
+   * @type {ActivationStrategy}
+   */
+  @bindable activationStrategy: ActivationStrategy = ActivationStrategy.InvokeLifecycle;
 
   /**
    * SwapOrder to control the swapping order of the custom element's view.
@@ -226,7 +250,7 @@ function processChanges(composer: Compose) {
   const changes = composer.changes;
   composer.changes = Object.create(null);
 
-  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes)) {
+  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes) && composer.activationStrategy !== ActivationStrategy.Replace) {
     // just try to activate the current view model
     composer.pendingTask = tryActivateViewModel(composer.currentViewModel, changes.model);
     if (!composer.pendingTask) { return; }
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -250,7 +250,7 @@ function processChanges(composer: Compose) {
   const changes = composer.changes;
   composer.changes = Object.create(null);
 
-  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes) && composer.activationStrategy !== ActivationStrategy.Replace) {
+  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes) && determineActivationStrategy(composer) !== ActivationStrategy.Replace) {
     // just try to activate the current view model
     composer.pendingTask = tryActivateViewModel(composer.currentViewModel, changes.model);
     if (!composer.pendingTask) { return; }
@@ -297,4 +297,13 @@ function requestUpdate(composer: Compose) {
     composer.updateRequested = false;
     processChanges(composer);
   });
+}
+
+function determineActivationStrategy(composer: Compose) {
+  let activationStrategy = composer.activationStrategy;
+  const vm = composer.currentViewModel;
+  if (vm && typeof vm.determineActivationStrategy === 'function') {
+    activationStrategy = vm.determineActivationStrategy();
+  }
+  return activationStrategy;
 }
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -4,7 +4,7 @@ import { TaskQueue } from 'aurelia-task-queue';
 import { bindable, CompositionContext, CompositionEngine, customElement, noView, View, ViewResources, ViewSlot } from 'aurelia-templating';
 
 /**
- * Available activation strategies for the view and view-model bound to compose
+ * Available activation strategies for the view and view-model bound to `<compose/>` element
  *
  * @export
  * @enum {string}
@@ -250,11 +250,7 @@ function processChanges(composer: Compose) {
   const changes = composer.changes;
   composer.changes = Object.create(null);
 
-  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes) && determineActivationStrategy(composer) !== ActivationStrategy.Replace) {
-    // just try to activate the current view model
-    composer.pendingTask = tryActivateViewModel(composer.currentViewModel, changes.model);
-    if (!composer.pendingTask) { return; }
-  } else {
+  if (needsReInitialization(composer, changes)) {
     // init context
     let instruction = {
       view: composer.view,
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
Makes sense to me. @bigopon Can you review as well?

EisenbergEffect:
My only wish is that we could somehow have the same lifecycle enum for templating and router, but I’m not sure how we can do that at this point. If the names and values are the same, then that’s at least a reasonable compromise. We could run into the issue where someone uses the third routing option with compose though.

Sayan751:
It seems that in aurelia e2e-tests package there is a [deprecated dependency on `@babel/polyfill`](https://github.com/aurelia/e2e-tests/blob/master/aurelia_project/aurelia.json#L48) which is causing the CI pipeline failure. Tested locally, removing that line, and the tasks executed without failure.

StrahilKazlachev:
@EisenbergEffect there was already a discussion, of sort, with @fkleuver about having the same names everywhere:
- https://github.com/aurelia/templating-resources/issues/381#issuecomment-506050865
- https://github.com/aurelia/templating-resources/issues/381#issuecomment-506420697

I do agree with him that it's confusing and you also point to this with:
> We could run into the issue where someone uses the third routing option with compose though.

Still, it would probably be better to go this way, having same name life-cycle "phases", since it's already a thing for *vCurrent*.

EisenbergEffect:
It's fine with me. We can unify this better in vNext. Since @bigopon is our most knowledgeable vCurrent templating team member, if he can take a quick look and give a thumbs up, I'm ok to merge this in. I was actually planning to do a a release of this library today, so if we can get one more review soon, I can get this out pretty quick. I'm happy to delay the release another day to get this if we don't think it will take much longer. After this is out, I'm releasing the new script library as well. I just wanted to get this stuff in there :)

src/compose.ts
```diff
@@ -226,7 +250,7 @@ function processChanges(composer: Compose) {
   const changes = composer.changes;
   composer.changes = Object.create(null);
 
-  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes)) {
+  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes) && composer.activationStrategy !== ActivationStrategy.Replace) {
```
bigopon: Can we do it via something like `resolve/determine ActivationStrategy`? And the exisiting change would be merged into that better, as in we normalize everything in a common handler path?

Sayan751: You mean to have a `determineActivationStrategy` hook in the bound VM, like it is done for the routed VMs? 

That actually crossed my mind and it should also be equally easy to implement. Only downside is that every bound VM that needs to be `replace`d should implement this hook as compared to applying a single attribute binding on the instance of `compose`. This might be annoying if the set of bound VMs for a single `compose` instance is considerably big. However, that can also be ameliorated with the help of inheritance.

If you like the trade-off, I am open to change the implementation.

EisenbergEffect: Probably we should have both. Like in routing, you can specify the strategy on the route itself, or you can implement the hook in the view model. Does that make sense?

Sayan751: That makes perfect sense. It offers more flexibility. 

I will make the change.

Sayan751: @bigopon requesting a review for the changes related to the `determineActivationStrategy` hook.

src/compose.ts
```diff
@@ -3,7 +3,22 @@ import { DOM } from 'aurelia-pal';
 import { TaskQueue } from 'aurelia-task-queue';
 import { bindable, CompositionContext, CompositionEngine, customElement, noView, View, ViewResources, ViewSlot } from 'aurelia-templating';
 
-
+/**
+ * Available activation strategies for the view and view-model bound to compose
```
bigopon:
```suggestion
 * Available activation strategies for the view and view-model bound to <compose/> element
```

bigopon: the suggestion is to highlight the usage of this, to avoid confusion between this and router `ActivationStrategy` when using auto import

src/compose.ts
```diff
@@ -274,3 +298,12 @@ function requestUpdate(composer: Compose) {
     processChanges(composer);
   });
 }
+
+function determineActivationStrategy(composer: Compose) {
+  let activationStrategy = composer.activationStrategy;
+  const vm = composer.currentViewModel;
+  if (vm && typeof vm.determineActivationStrategy === 'function') {
+    activationStrategy = vm.determineActivationStrategy();
```
bigopon: What if devs make a typo in `determineActivationStrategy` return? Do we normalize it or throw? At the moment, it doesn't handle it at all. I'd love to see a test for whatever the solution we come up with for this.
Thoughts?
cc @EisenbergEffect @fkleuver

src/compose.ts
```diff
@@ -226,7 +250,7 @@ function processChanges(composer: Compose) {
   const changes = composer.changes;
   composer.changes = Object.create(null);
 
-  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes)) {
+  if (!('view' in changes) && !('viewModel' in changes) && ('model' in changes) && determineActivationStrategy(composer) !== ActivationStrategy.Replace) {
```
bigopon: it feels to me the whole `!('view' in changes) && !('viewModel' in changes) && ('model' in changes)` should be in determine activation strategy as well, or abstracted to a separate function. But it just feeling, feel free to ignore this 😀

bigopon:
Looks good so far, I left a few comments. And I love the comments ❤️

src/compose.ts
Sayan751: For one, it is expected that devs use the `export`ed enum. However, if they make any mistake, then it will not be the `ActivationStrategy.Replace`, and the default strategy will come to play. Even if there comes 3rd strategy in future, it should not be a problem as long as it is matched against the enum values; and if it a no-match the default strategy will be used. Is an explicit error required in this case?

src/compose.ts
Sayan751: Actually that can be done. This stuff can be put in `determineActivationStrategy`, which normalizes the whole thing to a binary decision, activate or instantiate.

src/compose.ts
StrahilKazlachev: Does the router throw on unknown strategy?

bigopon: No it does not, as it check very differently 
https://github.com/aurelia/router/blob/893b768f01aea842ee57db4222e66aa572f24404/src/navigation-plan.ts#L123

https://github.com/aurelia/router/blob/893b768f01aea842ee57db4222e66aa572f24404/src/navigation-plan.ts#L142

@Sayan751 I forgot this `&& determineActivationStrategy(composer) !== ActivationStrategy.Replace` part when reviewing this bit. So I think we are good. But maybe let's add a test or two returning the wrong thing to solidify the logic here?

src/compose.ts
Sayan751: Done

src/compose.ts
Sayan751: Added negative tests

src/compose.ts
Sayan751: Refactored the implementation. Hopefully the changes make more sense now.

Sayan751:
@bigopon I have added some more tests for unknown values. Hope it is not much of a problem, as you have already reviewed. However, this should conclude the changes from my end.

src/compose.ts
```diff
@@ -274,3 +298,15 @@ function requestUpdate(composer: Compose) {
     processChanges(composer);
   });
 }
+
+function needsReInitialization(composer: Compose, changes: any) {
```
bigopon: Looks & feels much better 👍

test/compose.integration.spec.ts
```diff
@@ -81,6 +81,102 @@ describe('compose.integration.spec.ts', () => {
...
+    const oldModel = compose.model;
+    compose.modelChanged({}, oldModel);
+
+    taskQueue.queueMicroTask(() => {
+      expect(activationCount).toBe(2, 'activation count === 2');
+      component.dispose();
+    });
+  });
+
+  [1, true, 'chaos', new Date(1900, 1, 1), {}].map((strategy) =>
```
bigopon: 😁

test/compose.integration.spec.ts
```diff
@@ -81,6 +81,102 @@ describe('compose.integration.spec.ts', () => {
     component.dispose();
   });
 
+  it('works with determineActivationStrategy() - replace', async () => {
+    const { component, compose } = await bootstrapCompose(
+      `<compose view-model.bind="viewModel"></compose>`,
+      {
+        viewModel: class {
```
bigopon: just a note: it can also be an object, i.e: instance of this anonymous class: `new class { ...}`

bigopon:
It seems the ci is failing because of e2e repo tsconfig file is malformed. Irrelevant to this so I think it's safe to merge.
cc @fkleuver @EisenbergEffect

PR #383, opened 2019-07-07
WIP: feat(plugin,show,hide): support configurable hide-style

raugustinus:
I've added support for a configurable hide-style. We wanted this feature in support of the CSP configuration we are using which doesn't allow for inline-style elements. This was originally mentioned in this issue : aurelia/framework#54


raugustinus:
Changed the title to WIP:. Need to fix at least the package.json and package-lock.json and see why the ci build is failing.

bigopon:
Hmm I'm not sure why `@babel/polyfills` is needed here. This repo is no longer using babel 🤔

PR #385, opened 2019-07-10
Allow pug file extensions in html-resource-plugin

qraynaud:
I'm using pug & webpack and found out that
```html
<require from="resources/elements/simple-component.pug"></require>
```
was not working.

After looking into the codebase I found out that this seems to fix the issue though I'm not sure it's the proper way to do this.

Ideally I would have loved to find a way to make this configurable in `main.js` and even the webpack aurelia plugin `viewExtensions` option to be included in environment and looked for by the generated `main.js` file. Maybe with something like :

```js
export function configure(aurelia) {
  aurelia.use
    .standardConfiguration()
    .plugin(PLATFORM.moduleName('aurelia-validation'))
    .plugin(PLATFORM.moduleName('aurelia-animator-css'))
    .feature(PLATFORM.moduleName('resources/index'))
  ;

  if (environment.viewExtensions) {
    aurelia.setViewExtensions(environment.viewExtensions);
  }
}
```

But it's too big a change for me at this point. And I think it would have impacts cross repositories so it's better to get things done by smaller steps I suppose.

Hope this helps!

### DIFF
--- a/src/html-resource-plugin.ts
+++ b/src/html-resource-plugin.ts
@@ -1,30 +1,35 @@
 import {ViewEngine} from 'aurelia-templating';
 import {_createDynamicElement} from './dynamic-element';
 
+const VIEW_EXTENSIONS = ['html', 'jade', 'pug'];
+const VIEW_REGEXP = new RegExp('([^\\/^\\?]+)\\.(' + VIEW_EXTENSIONS.join('|') + ')', 'i');
+
 export function getElementName(address) {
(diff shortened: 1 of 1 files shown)


qraynaud:
I rebased my PR hoping it would fix the tests failing. I'm glad it did ;)

If you need me to improve my code over this, I'll be glad to do so. Just point me in the right direction and I'll do it.

I have been running this patch in production for over a month and it seems stable enough. I hope this helps.

PR #386, opened 2019-08-15
fix(all): change package.json module from es2015 back to native-modules

3cp:


3cp:
@bigopon one more.

bigopon:
This should also be good. Generally this change in our repo will not aurelia script anymore.
Thanks @3cp

3cp:
@EisenbergEffect

PR #391, opened 2020-01-16
fix(repeat): use utility fns for ie11 compat

### LINKED ISSUE #390: Capture and remove matcher binding breaks IE 11

* **Library Version:**
1.12.0

* **Operating System:**
Windows 10

* **Node Version:**
12.3.1

* **NPM Version:**
6.9.0

Aurelia CLI 1.2.2 with requireJS

* **Browser:**
IE 11

* **Language:**
ESNext

**Current behavior:**
Commit a5796443016d2c87269dd05801c057fb45e5dc44 introduced a breaking change for IE 11 
lines https://github.com/aurelia/templating-resources/blob/d346a817157960d54098a9193fc58c264ceedca6/src/repeat.ts#L289
and https://github.com/aurelia/templating-resources/blob/d346a817157960d54098a9193fc58c264ceedca6/src/repeat.ts#L295
require polyfills for documentFragment.children and documentFragment.firstElementChild respectively.

Without them in my case this caused console errors and the app failed to load.

App runs in IE 11

Maintain IE 11 compatibility

* **Additional Info**
For my app I added the polyfills from MDN:
https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children
```
// Overwrites native 'children' prototype.
// Adds Document & DocumentFragment support for IE9 & Safari.
// Returns array instead of HTMLCollection.
;(function(constructor) {
  if (constructor &&
    constructor.prototype &&
    constructor.prototype.children == null) {
    Object.defineProperty(constructor.prototype, 'children', {
      get: function() {
        var i = 0, node, nodes = this.childNodes, children = [];
        while (node = nodes[i++]) {
          if (node.nodeType === 1) {
            children.push(node);
          }
        }
        return children;
      }
    });
  }
})(window.Node || window.Element);
```
https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/firstElementChild
```
// Overwrites native 'firstElementChild' prototype.
// Adds Document & DocumentFragment support for IE9 & Safari.
;(function(constructor) {
    if (constructor &&
        constructor.prototype &&
        constructor.prototype.firstElementChild == null) {
        Object.defineProperty(constructor.prototype, 'firstElementChild', {
            get: function() {
                var node, nodes = this.childNodes, i = 0;
                while (node = nodes[i++]) {
                    if (node.nodeType === 1) {
                        return node;
                    }
                }
                return null;
            }
        });
    }
})(window.Node || window.Element);
```

bigopon:
closes #390 

cc @AquilaSands @EisenbergEffect

### DIFF
--- a/src/repeat.ts
+++ b/src/repeat.ts
@@ -286,13 +286,13 @@ export class Repeat extends AbstractRepeater {
       // if the template has more than 1 immediate child element
       // it's a repeat put on a <template/> element
       // not valid for matcher binding
-      if (template.children.length > 1) {
+      if (getChildrenCount(template) > 1) {
         return undefined;
       }
       // if the root element does not have any instruction
       // it means there's no matcher binding
       // no need to do any further work
-      const repeatedElement = template.firstElementChild;
+      const repeatedElement = getFirstElementChild(template);
       if (!repeatedElement.hasAttribute('au-target-id')) {
         return undefined;
       }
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@bigopon Before I merge, can you check the PAL to make sure we don't have some of these utilities already. It seems a bit familiar to me but it's been a while since I looked at that.

bigopon:
@EisenbergEffect we have `nextElementSibling`, but not `firstElementChild` and `childElementCount`

PR #396, opened 2020-09-06
fix(repeat): extract matcher once

bigopon:
Thanks to @Seebiscuit for the ping at https://github.com/aurelia/templating-resources/issues/295

cc @EisenbergEffect @fkleuver

### DIFF
--- a/src/repeat.ts
+++ b/src/repeat.ts
@@ -22,6 +22,8 @@ import {
 import {viewsRequireLifecycle} from './analyze-view-factory';
 import {AbstractRepeater} from './abstract-repeater';
 
+const matcherExtractionMarker = '__marker_extracted__';
+
 /**
  * Binding to iterate over iterable objects (Array, Map and Number) to genereate a template for each iteration.
  */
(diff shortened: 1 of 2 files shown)


src/repeat.ts
```diff
@@ -22,6 +22,8 @@ import {
 import {viewsRequireLifecycle} from './analyze-view-factory';
 import {AbstractRepeater} from './abstract-repeater';
 
+const matcherExtractionMarker = '__marker_extracted__';
```
Seebiscuit, line 25: Is there any downside to using a `Symbol` for the `matcherExtractionMarker ` property name? `Symbol` props are not enumerable, but that shouldn't be an issue here, since `Repeat` is not meant to be serialized (?).

bigopon: In this case, we are adding it on an instruction, and that's quite hidden away from apps, I think we are safe. For symbol vs string in general, I think it's about what's (probably?) more performant, easy to use (Aurelia polyfills does come with Symbol, but it's not used anywhere), so one can run an Aurelia app without having to polyfill a lot of stuff. This is true for all parts of our v1 core modules, we are not using Symbols.

Seebiscuit: Excellent point! I didn't realize there was no IE 11 support. Nice catch 💪

PR #429, opened 2023-09-08
export ComposeActivationStrategy closes #427

### LINKED ISSUE #427: [Types] compose activation strategy
**Library Version:**
1.14.3

as we can see from this PR https://github.com/aurelia/templating-resources/pull/382
we have the ability to control the activation strategy of the `<compose />`  element in two ways.
* by passing a bindable property
* by implementing `determineActivationStrategy` hook on the VM.

**Current behavior:**
* the enum `ActivationStrategy` (different from the one declared in `aurelia-router`) is not exported at all - making it hard to use in TS applications. (we need to use the strings litteral instead - or use the wrong enum).

other issue with the enum is briefly mentioned here: https://github.com/aurelia/templating-resources/pull/382#issuecomment-508168710
I think the best way to deal with it - is renaming the enum while exporting it to `ComposeActivationStrategy`

* exporting the `ActivationStrategy` enum using a different name: `ComposeActivationStrategy`.

I'll be working in the near future on a PR for this.
but I wanted to know if the approach I want to go is acceptable. (exporting the activation strategy as a different name)

avrahamcool:


bigopon:
hmmm simply merging this could be quite an issue for existing app. We may not be able to just do this. We can export an extra name though. Can you help change to an additional name instead @avrahamcool ?

avrahamcool:
hi @bigopon,

all of the previous exports remain untouched. [the `ActivationStrategy` enum was not exported before]
and I do export it as a deferent name (in order to not have a clash of name with the `ActivationStrategy` from router).
so how can this break existing apps?

I'm probably missing something here.

bigopon:
@avrahamcool thanks. I thought we were exporting it via `*` in the aurelia framework package, but it's not the case. This is good. Please ignore my previous comment.

bigopon:
The changes have been released in v 1.15.0
Thanks @avrahamcool

